PR #114, opened 2014-08-30
Added getRetryQueueSize 

fmmr:
to be able to monitor the retryqueue from outside this class. Also added getQueueSize for symmetry.


jchambers:
@fmmr I think this is a great suggestion, and we really appreciate the pull request. I think we're going to pass on this specific implementation for now in favor of doing something more robust (JMX, maybe?) as described in #175.

Thanks kindly for the contribution!

PR #121, opened 2014-09-11
Expose writable connection count

sigerber:
This change would allow clients using the PushMonitor to ascertain whether there are currently any usable connections to APNS. This information is useful for upstream service monitoring as it is not possible to reliably figure this out only by listening for failed connections.


jchambers:
@sigerber It sounds like there's wider demand for a feature like this, but I think we're going to pass on this specific implementation for now in favor of doing something more robust (JMX, maybe?) as described in #175.

Thanks kindly for the contribution!

PR #125, opened 2014-09-25
Abstract connection base class

jchambers:
As foretold by ~~prophecy~~ #124, this adds an abstract base class (`ApnsConnection`) for `PushNotificationConnection` (formerly `ApnsConnection`) and `FeedbackConnection`. The key advantages here are:

- De-duplication of connection/disconnection code
- Reusability/abstraction of unit tests
- Abstraction of common configuration options (see #126)
- Semantic clarity (i.e. reflecting in code the reality that these connections are very similar)
- Paying down technical debt on feedback connections (though this was mostly covered in #124)

~~The diff here is a bit noisier than it really needs to be since it's based off of the `feedback_service_listener` branch. When #124 gets merged, this will be much more readable. In the meantime, [comparing directly against  `feedback_service_listener`](https://github.com/relayrides/pushy/compare/feedback_service_listener...abstract_connection_base_class) is probably the way to go.~~

----

TODO:

- [ ] Go through the docs and update references to `ApnsConnection`
- [ ] Add docs for new interfaces


jchambers:
Closing temporarily; enough has changed in the old `ApnsConnection` that we'll need to do serious work to catch up here.

PR #126, opened 2014-09-25
Add optional connect/SSL handshake timeouts

jchambers:
This adds optional timeouts for establishing connections and completing an SSL handshake.


jchambers:
Closing temporarily; we'll need to do some pretty significant remediation work on the abstract base class branch in light of other changes.

PR #148, opened 2015-02-04
Adding support for title and its localized version - new in iOS 8.2

matzew:
The new `title`, `title-loc-key` and `title-loc-args` are described in here:
https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/ApplePushService.html

Especially interesting for Apple Watch:

* _"On Apple Watch, the title key alerts the user to the new request."_
* _"Apple Watch displays this string as part of the notification interface. This string is displayed only briefly and should be crafted so that it can be understood quickly. This key was added in iOS 8.2."_

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
@@ -42,6 +42,9 @@ public class ApnsPayloadBuilder {
 	private String alertBody = null;
 	private String localizedAlertKey = null;
 	private String[] localizedAlertArguments = null;
+	private String alertTitle = null;
+	private String localizedAlertTitleKey = null;
+	private String[] localizedAlertTitleArguments = null;
 	private String launchImageFileName = null;
 	private boolean showActionButton = true;
 	private String localizedActionButtonKey = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -128,6 +134,47 @@ public ApnsPayloadBuilder setLocalizedAlertMessage(final String localizedAlertKe
 	}
 
 	/**
+	 * <p>Sets a short description of the notification purpose. The Apple Watch will display this as part of the notification.
+	 * The string is visible for a short amount of time and should be crafted to be easy to understand.
```
jchambers: Minor, but it looks like you're missing a closing `</p>` tag here.

Also, rather than paraphrasing the docs, I think it makes more sense to quote directly (with attribution!) like we do for `setLaunchImage`. I know I haven't been super consistent on this point and apologize for that, but it seems like a good direction to go for new docs.

jchambers:
@matzew Looks good to me, except for that one minor documentation comment. Out of curiosity, how is it that you notice these changes? I'm subscribed to everything I can find from Apple, but stuff like this still catches me by surprise.

Also, it's bizarre to me that they updated the docs for this new stuff, but *still* haven't updated the docs to reflect the changes from WWDC 2014.

matzew:
@jchambers yo! updated the doc for setAlertTitle(). I didn't notice myself - I was pointed to it by a user of java-apns :-)

jchambers:
Looks good to me. Thanks, as always, for the contribution!

PR #152, opened 2015-02-19
Known-bad rejection handling (fixes #149)

### LINKED ISSUE #149: failed to find rejected notification
Found a new error log in our system:
```
failed to find rejected notification with sequence number 0 (buffer has range null to null);
 this may mean the sent notification buffer is too small. Please report this as a bug.
```

U-238:
This is a bugfix for #149. If the PushManager has sent the known-bad packet, and we then get a rejection message regarding sequence number 0, we will assume it was the known-bad packet and handle it accordingly.

Given that the first message we send has sequence number 1, it is unlikely that a legitimate rejection message for sequence number 0 would ever be received, unless we had sent more than 2^32 messages (about 4 billion) in one session. Nonetheless, the behaviour of assuming that sequence number 0 is the known-bad packet only applies *after* the known-bad packet has been sent.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -281,13 +281,17 @@ public class ApnsConnection<T extends ApnsPushNotification> {
 
 			// Regardless of the cause, we ALWAYS want to notify listeners that some sent notifications were not
 			// processed by the gateway (assuming there are some such notifications).
-			final Collection<T> unprocessedNotifications =
-					this.apnsConnection.sentNotificationBuffer.getAllNotificationsAfterSequenceNumber(
-							rejectedNotification.getSequenceNumber());
+			// Unless the sequence number was erroneously reported as 0, in which case we don't know the actual
+			// sequence number, and can't determine what was sent after the bad notification.
+			if (rejectedNotification.getSequenceNumber() != 0) {
+				final Collection<T> unprocessedNotifications =
+						this.apnsConnection.sentNotificationBuffer.getAllNotificationsAfterSequenceNumber(
+								rejectedNotification.getSequenceNumber());
 
-			if (!unprocessedNotifications.isEmpty()) {
-				if (this.apnsConnection.listener != null) {
-					this.apnsConnection.listener.handleUnprocessedNotifications(this.apnsConnection, unprocessedNotifications);
+				if (!unprocessedNotifications.isEmpty()) {
+					if (this.apnsConnection.listener != null) {
+						this.apnsConnection.listener.handleUnprocessedNotifications(this.apnsConnection, unprocessedNotifications);
+					}
 				}
 			}
 
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -254,7 +254,8 @@ public class ApnsConnection<T extends ApnsPushNotification> {
 			this.apnsConnection.sentNotificationBuffer.clearNotificationsBeforeSequenceNumber(rejectedNotification.getSequenceNumber());
 
 			final boolean isKnownBadRejection = this.apnsConnection.shutdownNotification != null &&
-					rejectedNotification.getSequenceNumber() == this.apnsConnection.shutdownNotification.getSequenceNumber();
+					(rejectedNotification.getSequenceNumber() == this.apnsConnection.shutdownNotification.getSequenceNumber()
+					|| rejectedNotification.getSequenceNumber() == 0);
 
 			// We only want to notify listeners of an actual rejection if something actually went wrong. We don't want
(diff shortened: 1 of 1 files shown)


src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
```diff
@@ -254,7 +254,8 @@ protected void channelRead0(final ChannelHandlerContext context, final RejectedN
 			this.apnsConnection.sentNotificationBuffer.clearNotificationsBeforeSequenceNumber(rejectedNotification.getSequenceNumber());
 
 			final boolean isKnownBadRejection = this.apnsConnection.shutdownNotification != null &&
-					rejectedNotification.getSequenceNumber() == this.apnsConnection.shutdownNotification.getSequenceNumber();
+					(rejectedNotification.getSequenceNumber() == this.apnsConnection.shutdownNotification.getSequenceNumber()
+					|| rejectedNotification.getSequenceNumber() == 0);
```
jchambers: It's probably unnecessary caution, but it might be good to check the rejection reason here, too:

```java
// ...
|| (rejectedNotification.getSequenceNumber() == 0 && RejectedNotificationReason.MISSING_TOKEN.equals(rejectedNotification.getReason()))
```

src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
```diff
@@ -280,13 +281,17 @@ protected void channelRead0(final ChannelHandlerContext context, final RejectedN
...
-			final Collection<T> unprocessedNotifications =
-					this.apnsConnection.sentNotificationBuffer.getAllNotificationsAfterSequenceNumber(
-							rejectedNotification.getSequenceNumber());
-
-			if (!unprocessedNotifications.isEmpty()) {
-				if (this.apnsConnection.listener != null) {
-					this.apnsConnection.listener.handleUnprocessedNotifications(this.apnsConnection, unprocessedNotifications);
+			// Unless the sequence number was erroneously reported as 0, in which case we don't know the actual
+			// sequence number, and can't determine what was sent after the bad notification.
+			if (rejectedNotification.getSequenceNumber() != 0) {
```
jchambers: I think this is the best thing we can do here, but let's pause here for a moment of arm-flailing while we consider the implications of this change. That this problem exists at all tells us that, even after the gateway has rejected a notification and is preparing to close, *it will still accept and send notifications received before the connection actually closes.*

I'm preeeeeeetty sure that clobbers what we had previously considered to be a guarantee about the state of a connection and its notifications after a notification is rejected. It seems like this leaves us between a rock and a hard place when it comes to post-rejection notifications; we can either send them again (which risks double-sending) or decide not to re-transmit (which risks never sending at all).

Uuuuuuuuuugh.

jchambers: Actually, one more thought—do we want to check the rejection reason here, too? Wrapping to `0` is, as we've all already acknowledged, really unlikely, but checking the reason might make it clearer what's going on (i.e. that we're working around a weird upstream issue).

jchambers:
@U-238 Thanks very much for tackling this! I think this is pretty much ready to merge, but it might be cool to add those extra rejection reason checks. I'll write a few bonus tests after this gets merged, and will plan on putting out a point release shortly thereafter.

src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
jchambers: Also also, I've filed http://openradar.appspot.com/radar?id=6372950655631360 with Apple about this issue.

U-238:
@jchambers I've added the code change you recommended. As for the guarantee about the state of a connection and its notifications after a notification is rejected, I think the lack of that guarantee is still preferable to having all push notifications resent again!

jchambers:
> I think the lack of that guarantee is still preferable to having all push notifications resent again!

I'm not sure I follow; could you rephrase?

In any case, I think this is good and will merge it. Thanks again! :metal:

U-238:
@jchambers Ok to rephrase: In your comments above you mentioned that this "clobbers what we had previously considered to be a guarantee about the state of a connection and its notifications after a notification is rejected". But it also fixes a bug where all the push notifications sent since startup would be sent again at shutdown (because it gets a rejection for sequence number 0, and therefore tries to resend everything in the buffer after number 0, i.e. everything).

U-238:
Thank you for accepting my pull request!

jchambers:
Ah! I see. Beg your pardon. I was complaining (needlessly) about the bug, not the fix--I absolutely agree that the proposed change is the right thing to do in the face of the *upstream issue* clobbering what we had previously considered to be a guarantee about state when a notification is rejected.

src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
jchambers: Good news! Looks like Apple fixed the delivery-after-rejection issue upstream!

PR #153, opened 2015-03-02
Format ExpiredToken token to readable string

runr:
Hi,

The ExpiredToken class uses Arrays.toString(token) to create a String version of the token byte array. I suggest using the TokenUtil to format it to a more "readable" string, making it easier to map the output of the expired tokens to actual device IDs.

Before: 
ExpiredToken [token=[0, 69, -18, 29, -118, 4, 90, 108, -1, -117, -36, -35, -55, -77, 84, -118, -57, 63, -114, 4, 77, -13, -106, 54, 105, -94, 72, -116, 64, -107, 65, -33], expiration=Mon Mar 02 09:08:23 CET 2015]

After:
ExpiredToken [token=0289822941142798c701b5450d50f51275b20c55a03ad895aee9ef0b29e94612, expiration=Sun Mar 01 22:41:08 CET 2015]

Please consider adding this to the next version of Pushy. Thanks for an otherwise excellent library!

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ExpiredToken.java
+++ b/src/main/java/com/relayrides/pushy/apns/ExpiredToken.java
@@ -24,6 +24,8 @@ package com.relayrides.pushy.apns;
 import java.util.Arrays;
 import java.util.Date;
 
+import com.relayrides.pushy.apns.util.TokenUtil;
+
 /**
  * <p>Represents a device token that the APN Feedback Service has reported as expired. According to Apple's
  * documentation:</p>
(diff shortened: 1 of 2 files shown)


jchambers:
Makes sense to me; thanks for the pull request!

I'm wondering if we should back off on the non-null requirements imposed by `TokenUtil`, though; those were added as a fail-fast kind of thing under the assumption that `TokenUtil` would only be called by Pushy users, and detecting `null` inputs early and aggressively might stave off some bugs in callers' code. Now that we're considering it as a way to interpret upstream data, though, I wonder if it makes sense to be a little more forgiving.

It would still be a bug (or at least a very big surprise) if we got an expiration notice with a null token from the feedback service, but throwing an exception from `toString` seems… tacky to me. Will contemplate and merge shortly.

runr:
Thanks for your response! 
Since ExpiredToken is immutable and the constructor will fail with a null token, you will never get a NPE from toString. Have added a test to demonstrate this.

jchambers:
> Since ExpiredToken is immutable and the constructor will fail with a null token…

Oh. Duh. A fine point!

> Have added a test to demonstrate this.

Thanks!

PR #157, opened 2015-03-12
Memory and synchronization improvements

Nowheresly:
Make static inner classes of FeedbackServiceConnection
Use of AtomicInteger for ApnsConnection.pendingWriteCount to ensure synchronization correctness.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: java
 jdk:
+    - oraclejdk8
     - oraclejdk7
     - openjdk7
     - openjdk6
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # pushy
 
+[![Build Status](https://travis-ci.org/relayrides/pushy.svg?branch=master)](https://travis-ci.org/relayrides/pushy)
+
 Pushy is a Java library for sending [APNs](http://developer.apple.com/library/mac/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Introduction.html) (iOS and OS X) push notifications. It is written and maintained by the engineers at [RelayRides](https://relayrides.com/) and is built on the [Netty framework](http://netty.io/).
 
 Pushy was created because we found that the other APNs libraries for Java simply didn't meet our needs in terms of performance or (especially) reliability. Pushy distinguishes itself from other libraries with several important features:


Nowheresly:
I can see why inner classes `FeedbackServiceConnection.ExpiredTokenDecoder` and `FeedbackServiceConnection.FeedbackClientHandler` could not be declared as static.
Accordingly the gain in memory consumption (two hidden pointers to the enclosing class) is very low.

The second modification is related to the field `ApnsConnection.pendingWriteCount`.
As far as I can see, it is sometimes read by using a lock on pendingWriteMonitor (see method `ApnsConnection.waitForPendingWritesToFinish`) and sometimes read/written using a lock on ApnsConnection (see `ApnsConnection.sendNotification`).
Since this field is not declared as `volatile`, I think this could lead to synchronization issues.
Any thoughts ?
In my PR, I choose to use an `AtomicInteger` instead of declaring it as `volatile`.

jmason:
+1 to using an AtomicInteger just to remove the potential risk.

I would suggest though that a cleaner fix might be to replace the pendingWriteCount and pendingWriteMonitor with a Semaphore, which has the semantics we want here.

jchambers:
> …a Semaphore, which has the semantics we want here.

Forgive me if I'm being dense here (it *is* Friday night), but could you elaborate on that point? I'm struggling to work out when we'd `acquire` and `release` (or maybe what our initial count would be?) to make that work.

Let me be clear, though: if there *is* a way to de-terrible-ize this thing with more standard concurrency tools, I'm all for it ;)

Nowheresly:
Maybe we could consider the use of semaphore to improve the current implementation as a different issue. Why not creating a new issue tagged as "enhancement" dedicated to this ?

jchambers:
> Why not creating a new issue tagged as "enhancement" dedicated to this?

We may wind up doing that in the end, but since we're already considering changing how we're handling pending writing counts, it seems reasonable to consider alternative approaches.

jmason:
actually, never mind me -- I've just taken a look, and yep, without a statically-sized pool of permits to start with I don't think a Semaphore would work here.  my mistake!

jchambers:
> without a statically-sized pool of permits to start with I don't think a Semaphore would work here.

Okay. Thanks for looking into it!

@Nowheresly This looks pretty good to me. Now that we have a `final AtomicInteger`, though, it seems like it might make sense to drop the existing `pendingWriteMonitor` and synchronize on the new `AtomicInteger` in `waitForPendingWritesToFinish` instead. Seem legit?

PR #164, opened 2015-04-25
Simplify immediate shutdown behavior

jchambers:
We previously had a really complicated way of closing connections immediately. The idea was that we'd add a task to the connection's event loop that would check on the connection's state, then make the most appropriate call as to how to shut things down. The goal was to avoid a situation where calling `close` (which I believe still doesn't happen on the event loop) could interrupt a read that would have otherwise told us about a rejected notification.

I think this change *does* expose us to that potential race condition again, but it's of sufficiently low probability that I don't think it justifies the added shutdown complexity.

Thoughts?

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -77,10 +77,8 @@ public class ApnsConnection<T extends ApnsPushNotification> {
 
 	private final String name;
 
-	private final Object channelRegistrationMonitor = new Object();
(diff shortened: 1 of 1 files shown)


AndreaLev:
Though it's a little risky, it seems like an acceptable tradeoff to me.

jchambers:
Yeah—I think the key thing is that it doesn't fundamentally change the guarantees we're making around shutdown behavior. Conversely, the added complexity from the old approach doesn't add any new guarantees, so I think simplicity is the way to go here.

PR #165, opened 2015-04-25
Schedule disconnection timeouts

jchambers:
Rather than adding a second idle state handler to the pipeline, we can just schedule an immediate disconnection after some delay. This seems a whole lot simpler to me.

Also, renamed "shutdown" to "disconnect" in the context of `ApnsConnections` for clarity.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -47,6 +47,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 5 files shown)


src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
```diff
@@ -326,6 +324,10 @@ public void channelInactive(final ChannelHandlerContext context) throws Exceptio
 					this.apnsConnection.listener.handleConnectionClosure(this.apnsConnection);
 				}
 			}
+
+			if (this.apnsConnection.gracefulShutdownTimeoutFuture != null) {
```
AndreaLev: Very minor: why does this still have `shutdown` when elsewhere that is renamed to `disconnect`?

AndreaLev:
Overall I think this is a good approach :+1:

src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
jchambers: Oops—just a goof on my part. Fixed in c4bab82.

jchambers:
Going to merge this even though Travis CI is reporting errors; I suspect it might be related to their [ongoing build environment update](http://www.traviscistatus.com/incidents/yk24wrkb4082).

PR #166, opened 2015-04-27
Don't notify listeners of connection closure until all writes have finished

jchambers:
This gets rid of the terrible terrible `waitForPendingWritesToFinish` method and generally cuts down on blocky badness. I'm still a little disappointed at how much hoop-jumping needs to happen here, but I feel much better now that it's entirely hidden from listeners/callers.

~~Depends on #164 and #165; apologies for the diff noise until those are merged. The commits to look at now are dbc04ab (which gets rid of the internal pending write counter) and 620cc77 (which gets write of `waitForPendingWritesToFinish` and wraps that all into more sensible timing for the `handleConnectionClosure` listener method).~~

This is easier to read with [whitespace-only changes hidden](https://github.com/relayrides/pushy/pull/166/files?w=1). Because we're not dealing with counters, we don't have as many anonymous `Runnables` lying around (yay!) and some stuff got un-indented as a result.


AndreaLev:
Looks good to me :+1:

jchambers:
Tests here are failing due to https://github.com/travis-ci/travis-ci/issues/3778.

PR #167, opened 2015-05-14
make more granular concurrent code

iryndin:
move out synchronized block code that need not be executed inside synchronized block in startNewConnection() method of PushManager

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/PushManager.java
+++ b/src/main/java/com/relayrides/pushy/apns/PushManager.java
@@ -824,13 +824,13 @@ public class PushManager<T extends ApnsPushNotification> implements ApnsConnecti
 	}
 
 	private void startNewConnection() {
+		final ApnsConnection<T> connection = new ApnsConnection<T>(this.environment, this.sslContext,
+				this.eventLoopGroup, this.configuration.getConnectionConfiguration(), this,
+				String.format("%s-connection-%d", this.name, this.connectionCounter++));
+
+		connection.connect();
+
 		synchronized (this.activeConnections) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/relayrides/pushy/apns/PushManager.java
```diff
@@ -824,13 +824,13 @@ public void handleUnprocessedNotifications(ApnsConnection<T> connection, Collect
 	}
 
 	private void startNewConnection() {
-		synchronized (this.activeConnections) {
-			final ApnsConnection<T> connection = new ApnsConnection<T>(this.environment, this.sslContext,
-					this.eventLoopGroup, this.configuration.getConnectionConfiguration(), this,
-					String.format("%s-connection-%d", this.name, this.connectionCounter++));
+		final ApnsConnection<T> connection = new ApnsConnection<T>(this.environment, this.sslContext,
+				this.eventLoopGroup, this.configuration.getConnectionConfiguration(), this,
+				String.format("%s-connection-%d", this.name, this.connectionCounter++));
```
eager: `++` is not thread-safe, so `connectionCounter` would need to be changed to an `AtomicInteger`.

jchambers: …which, honestly, probably makes more sense anyhow.

iryndin: Got it. Will fix and renew this pull-request

iryndin:
fixed, make connectionCounter to be AtomicInteger to increment it thread-safely outside synchronized block

jchambers:
Looks good. Thanks kindly!

PR #169, opened 2015-05-18
Encapsulate connection groups, add exponential back-off

jchambers:
This does a few things:

1. Connection lifecycle management gets moved out of `PushManager` and into `ApnsConnectionGroup`.
2. `ApnsConnectionGroup` opens connections by scheduling tasks on its event loop, which gives us a way to do exponential back-off.
3. Greatly simplified and untangled drain-before-shutdown behavior (no more weird interactions between the dispatch thread and the `PushManager`).

----

TODO:

- [x] Document the `ApnsConnectionGroup` class (9aca76a)
- [x] Resolve shutdown/timeout TODOs (eb4cc29)
- [x] Figure out how to test exponential back-off (eb4cc29, a45a45d)
- [x] Add or migrate tests for `ApnsConnectionGroup` (a45a45d)

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -496,6 +496,21 @@ public class ApnsConnection<T extends ApnsPushNotification> {
 	}
 
 	/**
+	 * Indicates whether this connection will write a push notification immediately. Any notifications sent when this
+	 * method returns {@code false} will be enqueued internally.
+	 *
+	 * @return {@code true} if this connection will write a push notification immediately or {@code false} if attempts
+	 * to send push notifications will be deferred
+	 */
+	public boolean isWritable() {
+		if (this.connectFuture == null || this.connectFuture.channel() == null) {
+			return false;
+		} else {
+			return this.connectFuture.channel().isWritable();
+		}
+	}
+
+	/**
 	 * Asynchronously sends a push notification to the connected APNs gateway. Successful notifications are
 	 * <strong>not</strong> acknowledged by the APNs gateway; failed attempts to write push notifications to the
 	 * outbound buffer and notification rejections are reported via this connection's listener.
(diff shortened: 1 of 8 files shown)


jmason:
Looks great!  it may be useful to limit the number of retry attempts if we continually fail without establishing a successful SSL connection to APNs.  The failure scenario here is that we're using an invalid or expired SSL cert; it'll never be able to connect, but will retry indefinitely.  (Sorry for the bike-shed comment ;) )

peter-yhc:
Hi Jon, do you know the time frame when this will be available as an official release?  We currently need the exponential back-off that you completed.

sunng87:
@jmason I had a PR #177 for this situation, I recommend you to check SSLException message in a listener to detect if the error is recoverable.

jmason:
@sunng87 actually, I've worked around it for now.  I found that expired or invalid SSL certs produce a ConnectionClosedException, which unfortunately can be produced in normal operation too -- so my workaround was to listen for the handleConnectionSuccess() callback and switch to an alternate, more lenient retry strategy once that is called.

jchambers:
> do you know the time frame when this will be available as an official release?

Sorry; I can't offer anything firm. I CAN say that we've just wrapped up an enormous project, though (RelayRides is now [Turo](http://turo.com/)!), and I expect to have more attention to devote to this in the very near future.

vkolotov:
We are looking forward to seeing these features in the master. We had an issue with connectivity to Apple APNS in production environment which caused overflooding our log files with netty and pushy logs about connection failures so we run out of disk space. For now we just enabled logback filter to filter out repetitive netty and pushy messages:

`<turboFilter class="ch.qos.logback.classic.turbo.DuplicateMessageFilter"/>`

jmason:
hey BTW, I have a nasty hack for people looking for this feature in the short term -- use a Guava MoreExecutors.sameThreadExecutor() as the listenerExecutorService, and you can then effectively block the main (sending) thread with sleeps in the handleFailedConnection() callback.  it'd be much nicer to have "official", clean support for backoff, though.

PS: Also worth noting that, as I mentioned above, we have found that different retry/backoff strategies are necessary for the "pre-connect" and "post-connect" stages; post-connect failures will be common due to how invalid tokens are handled, but pre-connect failures often indicate a configuration error (ie a bad ssl cert).

jchambers:
Closing this in favor of just doing HTTP/2 support (which will have exponential back-off).

PR #170, opened 2015-05-19
Fix ApnsPayloadBuilder alert body check when title is included (fixes #168)

### LINKED ISSUE #168: ApnsPayloadBuilder generates simple alert text instead of alert dictionary
The following codesnippet:

```java
final String payload = new ApnsPayloadBuilder()
        .setAlertBody("BODY")
        .setAlertTitle("TITLE")
        .buildWithDefaultMaximumLength();
```

results in this payload:

```json
{
  "aps": {
    "alert": "BODY"
  }
}
```

My expectation would be:

```json
{
  "aps": {
    "alert": {
      "body": "BODY",
      "title": "TITLE"
    }
  }
}
```

So, if I add any property (other than alert body) to the builder, and that property belongs into the alert dictionary, then the dictionary should actually be generated.

Possible fix: ApnsPayloadBuilder#shouldRepresentAlertAsString should additionally consider these properties:
alertTitle, localizedAlertKey and localizedAlertTitleKey

Best regards
Christian Charles

sunng87:
The current implement just ignored alert.title when deciding whether to set alert as string or dict.

Hoping to fix [#168]. Test included.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
@@ -519,6 +519,12 @@ public class ApnsPayloadBuilder {
 	 * should be represented as a dictionary
 	 */
 	private boolean shouldRepresentAlertAsString() {
-		return this.alertBody != null && this.launchImageFileName == null && this.showActionButton && this.localizedActionButtonKey == null;
+		return this.alertBody != null && this.launchImageFileName == null && this.showActionButton
+				&& this.localizedActionButtonKey == null
+				&& this.alertTitle == null
+				&& this.localizedAlertTitleKey == null
+				&& this.localizedAlertKey == null
+				&& this.localizedAlertArguments == null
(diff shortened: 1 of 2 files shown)


jchambers:
:metal:

Thanks kindly for the fix!

sunng87:
@jchambers Thanks for merging. 

Is it possible to create a release for this patch on 0.4.x branch? We have a customer waiting for this to go production. Also as pushy now has more and more users, perhaps it's time for us to setup development  and stable branch so we can fix fast and release fast.

jchambers:
> Is it possible to create a release for this patch on 0.4.x branch?

Yes, but it will have to wait for a few days.

> perhaps it's time for us to setup development and stable branch so we can fix fast and release fast.

The broad goal for us is that master is always stable and shippable. We tend to keep major new features in their own side branches (see #169, for example). I may be misunderstanding your suggestion, though—please let me know if I'm off the mark.

sunng87:
Thanks, Chambers! That works for me.

PR #174, opened 2015-05-24
Variable maximum payload lengths (fixes #151)

### LINKED ISSUE #151: Extended payloads only supported under iOS 8 and newer
Apple has, at last, updated [their docs](https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/ApplePushService.html#//apple_ref/doc/uid/TP40008194-CH100-SW9) to cover the expanded payloads announced at WWDC 2014. However, the docs describe a situation a little more nuanced than I think we're currently reflecting in our own docs and API design:

> In iOS 8 and later, the maximum size allowed for a notification payload is 2 kilobytes; Apple Push Notification service refuses any notification that exceeds this limit. (Prior to iOS 8 and in OS X, the maximum payload size is 256 bytes.)

We should certainly update our own docs and probably update the payload construction tools, too.

jchambers:
This change makes the maximum payload length for a push notification depend on the version of iOS running on the receiving device. Unfortunately, callers will need to keep track of the iOS version associated with APNs tokens on their own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,8 @@ final ApnsPayloadBuilder payloadBuilder = new ApnsPayloadBuilder();
 payloadBuilder.setAlertBody("Ring ring, Neo.");
 payloadBuilder.setSoundFileName("ring-ring.aiff");
 
-final String payload = payloadBuilder.buildWithDefaultMaximumLength();
+final String payload =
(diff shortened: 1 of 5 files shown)


jchambers:
Closing this for now; that latest revision of the APNs protocol claims that we can now do 4kB payloads to all devices regardless of OS version.

PR #177, opened 2015-06-01
(feat) Do not retry when connection rejected by a fatal reason.

sunng87:
This patch includes a breaking change to FailedConnectionListener.
Added a "fatal" argument indicates this exception is fatal and we
won't retry.

Some unrecoverable certificate issues are considered fatal for now, such as `certificate_revoked` or `certificate_expired`. These errors require users to recreate their PushManager with corrected SSLContext. Retrying just doesn't help.

One question left is do we need to shutdown PushManager for user in this situation or give it to user? WDYT?

edit: spelling and wording

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/FailedConnectionListener.java
+++ b/src/main/java/com/relayrides/pushy/apns/FailedConnectionListener.java
@@ -47,5 +47,5 @@ public interface FailedConnectionListener<T extends ApnsPushNotification> {
 	 * @param pushManager the push manager that failed to open a connection
 	 * @param cause the cause for the connection failure
 	 */
-	void handleFailedConnection(PushManager<? extends T> pushManager, Throwable cause);
+	void handleFailedConnection(PushManager<? extends T> pushManager, Throwable cause, boolean fatal);
 }
--- a/src/main/java/com/relayrides/pushy/apns/PushManager.java
+++ b/src/main/java/com/relayrides/pushy/apns/PushManager.java
@@ -21,6 +21,7 @@
 
 package com.relayrides.pushy.apns;
 
+import com.relayrides.pushy.apns.util.ExceptionUtil;
 import io.netty.channel.nio.NioEventLoopGroup;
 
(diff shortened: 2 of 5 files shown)


jchambers:
I like the idea of helping users identify fatal connection failures (we tell users to watch for fatal exceptions in the README, but don't provide much guidance on identifying them). I think the key thing here will be accurately identifying the exceptions that are actually fatal. With that in mind, I have a few questions:

1. How confident are you that fatal exceptions will always contain the word "fatal" in the message?
2. Are there cases where the word "fatal" appears in the message, but the exception is not actually fatal? I can imagine this happening if an exception, for example, had the term "non-fatal" in its message.
3. Will different JVMs or security providers produce different messages for the same exceptions?

sunng87:
The check is based our production observation. Currently, we only check message from SSLException. You are right, I will look into jdk documents to see if there's deterministic method for this check.

sunng87:
Sadly, SSLException doesn't expose any API to indicate whether it's fatal or not. Before we actually find a deterministic solution, I just changed the check to `e.getMessage().contains("Received fatal alert");`. Based on our observation this could catch most fatal SSL fatal issues.

If we don't stop PushManager on fatal alert, the connector will fall into a dead loop (connect -> rejected) which may causing issues in your system.

jchambers:
I'm hesitant to make an API-level guarantee based on undocumented, human-readable text. That said, I think you're absolutely right that callers should shut down their `PushManager` if one of these fatal exceptions happens, and I think you've definitely identified a useful way to check for fatal exceptions in *at least* one JVM/security provider combination.

Rather than baking this into the Pushy API, would it make sense to add it as a suggestion to the docs and example code?

sunng87:
Totally understood. We have added code to shutdown PushManager in our listener, by the custom check on SSLException. It is true that there's no deterministic way to know if it's fatal. We can add this to doc/readme as a best practice for now.

jchambers:
Cool. Thanks for your understanding, and also for the suggestion! For what it's worth, #169 should make the cost of connection cycling much, much lower.

PR #179, opened 2015-06-07
Even simpler channel closure!

jchambers:
I'm 99% sure this will resolve a race condition we've seen causing intermittent failures test failures in #169 and #174. Basically, instead of checking to see whether an operation is the last thing in the queue (which led to potential double closure notifications), we just add the closure notification to the end of the queue and trust that it will be the last thing to happen.

Seems waaaaaaaaay simpler than what we were doing before and less error-prone. The thing I don't like is that there isn't a documented guarantee from Netty about the order of things in 4.0 (though there is in 4.1).

Thoughts? @sunng87, for what it's worth, I think we need to resolve this before we can put out a release with the fix for #168.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -86,8 +86,6 @@ public class ApnsConnection<T extends ApnsPushNotification> {
 	// having an expired token) is vanishingly small.
 	private int sequenceNumber = 1;
 
-	private volatile ChannelFuture lastWriteFuture;
-
 	private int sendAttempts = 0;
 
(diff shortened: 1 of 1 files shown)


jchambers:
> What if an eventLoop is shared among multiple channels? Is this runnable task associated per-channel?

So multiple `Channels` can share an `EventLoopGroup`, but each individual channel gets its own `EventLoop` within that group. Further, one thread can have many event loops, but each event loop is always bound to a single thread. The model guarantees that each channel has its own "territory," and events in its loops will run in series (barring deliberate scheduling shenanigans).

sunng87:
Understood. That makes sense. Thanks for your explanation.

PR #190, opened 2015-08-19
Inject Executor instead of ExecutorService for listeners callbacks

magnus-gustafsson:
Since Pushy only should not shut down an provided ExecutorServices there is no need for it to actually be an ExecutorService instance. 
Providing an Executor is enough for Pushys use case.
This will make it easier to use Pushy in our managed environment since we can reuse our own Executors for its listeners callbacks.


jchambers:
Sorry to leave this for so long! Now that HTTP/2 is A Thing™ (see #213), I don't think we'll need listeners at all any more. I'm going to close this for now, but please accept our sincere thanks for the contribution!

Issue #196, opened 2015-09-28
Pushy continues enqueueing onto closed/poisoned connections


omarkilani:
I had a situation where pushy (latest master) continued to write onto a closed connection forever.

It would 'failed to write notification' with a cause of 'java.nio.channels.ClosedChannelException', over and over on that failed channel. The channel received an INVALID_TOKEN rejection prior to this.

I switched to using the new APNSConnectionGroup branch (encapsulate_connection_groups) but it still enqueues onto a closed channel:

```
TRACE [2015-09-28 23:02:29,807] [Thread-5] com.relayrides.pushy.apns.ApnsConnection - APNSPushManager-ConnectionGroup-3 sending SendableApnsPushNotification [sequenceNumber=1850, token=abc, payload={"aps":{"content-available":1}}, deliveryInvalidation=Tue Sep 29 23:02:29 UTC 2015]
TRACE [2015-09-28 23:02:29,808] [nioEventLoopGroup-2-4] com.relayrides.pushy.apns.ApnsConnection - APNSPushManager-ConnectionGroup-3 successfully wrote notification 1850
DEBUG [2015-09-28 23:02:29,904] [nioEventLoopGroup-2-4] com.relayrides.pushy.apns.ApnsConnection - APNs gateway rejected notification with sequence number 1850 from APNSPushManager-ConnectionGroup-3 (INVALID_TOKEN).
TRACE [2015-09-28 23:02:29,922] [nioEventLoopGroup-2-4] com.relayrides.pushy.apns.ApnsConnectionGroup - ApnsConnection [name=APNSPushManager-ConnectionGroup-3] rejected SimpleApnsPushNotification [token=[1, 2, 3], payload={"aps":{"content-available":1}}, invalidationTime=Tue Sep 29 23:02:29 UTC 2015, priority=IMMEDIATE]: INVALID_TOKEN
TRACE [2015-09-28 23:02:29,929] [nioEventLoopGroup-2-4] com.relayrides.pushy.apns.ApnsConnectionGroup - Connection closed: ApnsConnection [name=APNSPushManager-ConnectionGroup-3]
TRACE [2015-09-28 23:02:29,932] [nioEventLoopGroup-2-4] com.relayrides.pushy.apns.ApnsConnectionGroup - ApnsConnectionGroup [name=APNSPushManager-ConnectionGroup] will remove connection ApnsConnection [name=APNSPushManager-ConnectionGroup-3].
[...]

PR #197, opened 2015-09-29
Fixes for #196 - Pushy continues enqueueing onto closed/poisoned conn…

omarkilani:
Fixes for #196.


jchambers:
Good catch. As you've pointed out in a separate pull request, we're mostly focused on HTTP/2 for now. I'll make sure that the lessons from this pull request make it into the HTTP/2 implementation, though! Thanks very much for the contribution, and I'm sorry to have let it sit idle for so long!

Issue #184, opened 2015-06-17
HTTP/2 support


matzew:
Apple will have HTTP/2 API to send notification requests to APNs.

YAY! :tada: 

Here is a little summary of more details:
* request/reponse (aka stream) for every sent (more reliable (e.g. JSON reason for a ‘bad request’ or 410 if the token is invalid))
 * allows ‘instant’ feedback (no separate feedback service!), since details are on the HTTP/2 response
* multiplexing: multiple requests (to APNs) on a single connection
* binary
* simpler certificate handling: Just a single cert! (no separate for dev/prod, VOIP etc)
* 4KB size of payload (for all versions of iOS/OSX), but just on the new HTTP/2 API

The new HTTP/2 API for APNs will be available in “Summer 2015″ for the development environment and will be made available for production “later this year”. No exact dates were given.

Details: https://developer.apple.com/videos/wwdc/2015/?id=720

sunng87:
fuck yea!!!

jchambers:
What an unreasonably huge relief ;)

Pushy is going to be a much, much simpler piece of software once this happens.

jchambers:
I've asked how we can learn when the HTTP/2 docs become available, and I've been told that the best way to find out is to just keep monitoring developer.apple.com (I've asked for clarification as to whether the docs will appear in the main "Local and Remote Notification Programming Guide," or if they will appear in a separate document). I've been checking periodically, but please do post here if you see them before I do.

Thanks!

matzew:
sounds like a plan

On Tue, Jun 30, 2015 at 5:40 AM, Jon Chambers <notifications@github.com>
wrote:

> I've asked how we can learn when the HTTP/2 docs become available, and
> I've been told that the best way to find out is to just keep monitoring
> developer.apple.com (I've asked for clarification as to whether the docs
> will appear in the main "Local and Remote Notification Programming Guide,"
> or if they will appear in a separate document). I've been checking
> periodically, but please do post here if you see them before I do.
>
> Thanks!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/relayrides/pushy/issues/184#issuecomment-116934872>.
>



-- 
Matthias Wessendorf

blog: http://matthiaswessendorf.wordpress.com/
sessions: http://www.slideshare.net/mwessendorf
twitter: http://twitter.com/mwessendorf

matzew:
[...]

PR #204, opened 2015-10-28
Add method getExpiredTokens which return Future with results

iryndin:
Add method getExpiredTokens which return Future with results. Also added unit tests for it.


jchambers:
First and foremost, thank you very much for the contribution! I'm sorry to have taken so long to get to this, but I think we may not actually need to do this now that the feedback service is going away (see #184 and #213). I'm going to close this for now, but again, thank you very much!

Issue #142, opened 2014-11-20
Direct memory exhausted after sending many notifications


jchambers:
Most of the details so far have been posted to [a mailing list thread](https://groups.google.com/d/msg/pushy-apns/W3b24t1PkHY/QnU_QpqnYF0J), but to recap:

1. When sending notifications, we'll sometimes run out of direct memory when attempting to write a notification to a channel in `ApnsConnection`. We thought this was fixed when Netty 4.0.24 came out (see #138), but the issue still seems to persist.
2. Just sending zillions of notifications to a mock server results in stable memory usage, though the problem crops up in real-world usage. We suspect the problem is centered around connection closure/rejected notifications.
3. We don't run out of direct memory until many (millions, we think) of notifications have been sent under real-world conditions.
4. The problem appears to be that *something* in a Netty event loop is (a) persisting longer than it should ~~and (b) holding a reference to an `ApnsConnection`, which is causing a leak~~ (that last part doesn't make sense, since that *shouldn't* be taking up any direct memory).

jchambers:
Of interest: http://blog.csdn.net/jiangguilong2000/article/details/42297873

jchambers:
Oh, hey, @ylgrgyq, that's your blog post, isn't it? ;)

sunng87:
We confirm upgrading Netty 4.0.25 doesn't help. Also I'm afraid the direct memory leak only happens to secure connection. We have another application running on Netty and suffering the leak.

Still haven't got anything valuable for debugging.

ylgrgyq:
> Oh, hey, @ylgrgyq, that's your blog post, isn't it? ;)

haha, sorry, that's not me. That's a poor Chinese programmer who confront with the same problem as me. 
I wonder how do you find that blog and how do you know I'm a Chinese? Please don't tell me you conclude that from my poor English. ;)

ylgrgyq:
> We confirm upgrading Netty 4.0.25 doesn't help

I saw this issue was reopened. https://github.com/netty/netty/issues/3166

jchambers:
> I wonder how do you find that blog and how do you know I'm a Chinese?

Found it by searching (via Google) for other people having the same problem. I thought it was you because it's exactly the same problem/data you reported, but I guess not!

ylgrgyq:
Hi, it's me again.
[...]

PR #206, opened 2015-11-07
(fix) check whether channel is writable before writeAndFlush

ylgrgyq:
As we talked in this issue :https://github.com/relayrides/pushy/issues/142 that Pushy don't need to check whether the channel is writable before writeAndFlush to this channel because Pushy has writability listener which can do this job. I think it's very reasonable at that time. 

But I found a clue which may indicate that netty's writability listener can have a big lag behind the change time of the channel writability.  It means channel may already unwritable but the writability listener is not triggered for a long time. At that time, if we don't check the channel's writability we can write a lot of pending data far beyond the high water mark in netty. If the channel will recover eventually, that's all ok. But if the channel is broken, all the pending data wrote to this channel is wasted.

image

Please look at this picture. I got it from a heap dump opened by Eclipse Memory Analyzer. It's a detailed attributes information from a ChannelOutboundBuffer instance which belongs to an ApnsConnection. This ApnsConnection is unwritable and is removed from writableConnections queue in PushManager correctly. But in this picture you can see that unwritable is 1 but totalPendingSize is 1460866 (> 1MB). But the default value of the high water mark in netty is 64KB (we don't change it) which means (In my opinion) this channel has wrote a lot of data  to this channel before noticed that the channel is unwritable already.

By the way, I still think this issue has nothing to do with the memory leak mentioned in https://github.com/relayrides/pushy/issues/142

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -512,7 +512,7 @@ public class ApnsConnection<T extends ApnsPushNotification> {
 			throw new IllegalStateException(String.format("%s has not completed handshake.", this.name));
 		}
 
-		if (this.disconnectNotification == null) {
+		if (this.disconnectNotification == null && this.connectFuture.channel().isWritable()) {
 			final SendableApnsPushNotification<T> sendableNotification =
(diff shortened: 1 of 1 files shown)


jchambers:
I'm sorry to take so long to get to this! I think you're totally right. I'll merge this here and make sure we do the same thing in #213.

PR #207, opened 2015-11-09
Fixed task leak in ApnsConnection/gracefullyDisconnect

sunng87:
Hi @jchambers , we detected a leak of scheduled task during shutdown and this is the fix.

cc: @ylgrgyq


jchambers:
Oops. Good catch!

PR #209, opened 2015-12-02
(fix) donot retry gracefully disconnect on IOException

sunng87:
In our production we saw lots of WriteAndFlushTask that blocks event thread and the whole system stops working. 

My colleague @ylgrgyq and I found if issue like "Broken pipe" or "Connection
reset by peer" happens on `ApnsConnection`, we will fall into an infinite retry of GracefullyShutdown until it meets timeout. This might produce a lot of `WriteAndFlushTask` and block event threads, stop it to register new channels, then the system stops working totally.

And these `IOException`s, there is no chance to recovery by retry so we better close the connection immediately.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsConnection.java
@@ -42,6 +42,7 @@ import io.netty.handler.timeout.IdleStateHandler;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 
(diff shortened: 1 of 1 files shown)


jchambers:
Mercifully, we won't need to worry about this at all in the [HTTP/2 implementation](https://github.com/relayrides/pushy/pull/213) (or, at least, "graceful shutdown" will be a well-understood idea that's actually supported by the underlying framework). I'm going to close this for now, but thank you very much for the contribution! Sorry to let it sit for so long!

PR #212, opened 2015-12-04
+add successful listener was added

devhak2:
Hi.
When i work with this, i need some successful listener to accumulate success count.
So i added successful listener, but i know that is not real success cause we don't recognize whether APNs really  send this or not. However, i think this is useful for some people including me.
I hope this is reflected.
Please check this.


jchambers:
@devhak2 First and foremost, thank you for the contribution! I think it makes sense to pass on this pull request for now. Once #213 is ready, you'll be able to track the status of each individual notification.

Again, thank you!

Issue #189, opened 2015-08-12
PushManager.isShutDown() is not thread-safe


jmason:
hey folks --

PushManager.isShutDown() ( https://github.com/relayrides/pushy/blob/master/src/main/java/com/relayrides/pushy/apns/PushManager.java#L318-L327 ) has a thread-safety issue -- this.shutDownStarted is a boolean member of the PushManager object and is accessed without any lock.  There's therefore a risk that a caller in a different thread from the one calling shutdown() will see an incorrect value.

It should be easy enough to make shutDownStarted and shutDownFinished volatile, to avoid this risk.

I think similar thread-safety issues arise with dispatchThread (which is accessed by the public isStarted() API which is in turn called by requestExpiredTokens()), and dispatchThreadShouldContinue (which may be accessed by the dispatch thread while shutdown() is in progress).

jchambers:
Yikes; totally missed this one when it opened. Sorry, @jmason!

Yep; that all sounds sensible to me. Man am I looking forward to HTTP/2 support so we don't have to do all this stuff.

jmason:
great! lemme know if you want me to fire in a quick pull request.

jchambers:
This should be a non-issue now that we've moved to the HTTP/2-based APNs protocol (see #213). Closing for now; please leave a comment if this resurfaces later!

Issue #175, opened 2015-05-24
Expose important metrics


jchambers:
This kind of thing has come up in a number of different contexts. #114 and #121 both offer partial solutions, but I think the most robust way to go (as [suggested on the mailing list](https://groups.google.com/forum/#!searchin/pushy-apns/jmx/pushy-apns/mgBoaqUAwGk/qMymnt90O5wJ)) would be to use something like JMX or [Metrics](https://dropwizard.github.io/metrics/). Opening this issue as a placeholder/catch-all for this kind of request and for discussion of possible approaches.

jmason:
one approach for libs to expose metrics is to provide an API which freezes a snapshot of current metric state, and callers can then write a simple adapter to copy those to their desired metric backend.  Guava does this on their Cache objects: https://code.google.com/p/guava-libraries/wiki/CachesExplained#Statistics

jchambers:
Now that we've shipped HTTP/2 support, I think this will be much more straightforward. I've put out a [request for advice on best practices](https://groups.google.com/d/msg/metrics-user/gPdBrs2R_94/9td7_JLeFAAJ) as to how we'd manage metrics when there are lots of `ApnsClient`s in play and plan on tackling this as soon as I get an answer.

jmason:
I think it’d be a good idea not to tie Pushy to a specific metrics library
— expose an interface which callers can implement to receive metrics
callbacks.  Some people using Dropwizard Metrics, some use Datadog (via
statsd), some use statsd directly.

BTW, Hystrix is a good example of a production-quality library which
exposes metrics well, it may provide some ideas:
https://github.com/Netflix/Hystrix/wiki/Metrics-and-Monitoring-(up-to-1.4.x)
.  Their approach of having a “metrics-publisher” maven jar, which can be
instantiated and registered against the core Hystrix lib, works well, in my
experience.

Also, Marshall Pierce’s idea of specifying a string prefix for each set of
metrics is a good one — I use that internally in our code which logs
metrics (to graphite via Dropwizard Metrics).

—j.

jmason:
Here's something demoing the kind of thing I mean: https://github.com/relayrides/pushy/pull/222

jchambers:
> Here's something demoing the kind of thing I mean: #222

Awesome. Thank you! That answered a lot of the questions I had in my head, but hadn't found the time to articulate.

Looking at #222, it seems to me that there are two extremes of approaches we could take to exposing internal metrics:

[...]

PR #213, opened 2015-12-10
HTTP/2

jchambers:
Apple [announced at WWDC 2015](https://developer.apple.com/videos/play/wwdc2015-720/) that APNs will be moving to HTTP/2, and formally released [the specification](https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/Introduction.html) a few weeks ago. Switching to HTTP/2 promises to resolve many, many of the issues that had made life difficult in the past. Some key changes:

- We'll get responses for each individual notification, whether accepted or rejected
- The feedback service is gone, and we'll get an HTTP/410 with a timestamp if a token has expired (because the recipient has uninstalled the receiving app)
- Payload size has increased to 4kB

One really important thing to note: the system requirements for Pushy—really, any bit of Java that's trying to implement HTTP/2, regardless of the underlying framework—are going to get a bit more complicated. See https://github.com/netty/netty/issues/4522 for additional discussion.

TODO:

- [x] Update docs
  - [x] Update README
  - [x] Update API docs
  - [x] Document strategies for dealing with HTTP/2 system requirements
- [x] Check for leaks
- [x] Add support for OpenSSL
- [ ] Restore fancy build matrix in Travis CI
  - [ ] Add tests for OpenSSL (will depend on Netty 4.1.0.CR1 due to https://github.com/netty/netty/pull/4576)
- [x] Restore logging (67c907c)


jmason:
yay! good to see this starting. Are Apple offering a HTTP/2 API endpoint yet?

the deployment story for HTTP/2, with alpn-boot etc., is extremely painful -- I've just got a http2 service into production and it wasn't fun.  I actually wound up going with OkHttp instead of Netty, as I just couldn't get Netty interoperating with other HTTP/2 implementations (probably due to cipher problems) on java 7.

jchambers:
> Are Apple offering a HTTP/2 API endpoint yet?

No, but poking around a bit, it sounds like other people are having some success just talking to the endpoint Apple mentioned in their WWDC examples.

> …I just couldn't get Netty interoperating with other HTTP/2 implementations (probably due to cipher problems) on java 7.

Yeah—Java 7 won't cut the mustard on its own. HTTP/2 depends on ciphers that didn't get introduced until Java 8. I think there are three options for addressing the cipher issue:

- Use Java 8 or newer
- Use OpenSSL (bonus: I believe this also solves the ALPN dependency issue)
- Use Bouncy Castle or another provider

The Netty team is building up docs around the bonus system requirements for HTTP/2, and I'm planning on chipping in there as soon as I can get some quiet time to write.

jmason:
I've heard of #1 and #2 -- the BouncyCastle option is new to me. looking forward to seeing it progress ;)

sunng87:
Jetty is offering a alpn-boot package to patch some jdk classes and enables ALPN for jdk prior to 8:
http://www.eclipse.org/jetty/documentation/current/alpn-chapter.html

jchambers:
Hey, everybody! Looks like the updated APNs protocol has been published! See https://developer.apple.com/news/?id=12172015b for the announcement and https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/Introduction.html for the details.

jmason:
https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/APNsProviderAPI.html#//apple_ref/doc/uid/TP40008194-CH101-SW1 -- no more feedback service, a sane HTTP/2-based protocol -- looks so much better!

jchambers:
Folks, I think we're pretty close to ready here. I would describe what we have now as functionally complete; I think the main things to do now are checking for leaks with reference-counted things in Netty (I have no idea what I'm doing there) and then documenting everything.

It's worth noting that this looks *very very different* from previous versions of Pushy, but I think that's a very good thing. The `PushManager` class is gone, and is replaced by a much simpler `ApnsClient`. I'm especially curious what you all think there. Please let me know if you have any questions or concerns now that things are taking shape!

flozano:
Thanks for working on this! it looks great :+1: 

One risky suggestion - do you think some kind of "listenable" future (eg: Java8's CompletableFuture/CompletionStage, Guava's ListenableFuture... or even a custom simple interface) could be used instead of plain Future<V>? It helps a lot when writing really async code, in order to avoid polling for Future<?>'s completion status or - even worse - block on it.

I think the easiest would be using CompletableFuture from J8, as I see you removed older versions from the CI file (but not sure if that's intended to stay this way when releasing).

BTW Last commit (avoid netty-all) helps me a lot :+1:

flozano:
I'm asking because I'll definitely use this when you release it :o) but if it stays like this I will have to wrap around a polling layer for Future<?>, and I guess I'm not the only one doing async request pipelines these days...

jchambers:
> …do you think some kind of "listenable" future… could be used instead of plain `Future`?

Yes. My thought would be to use Netty's [`Future`](http://netty.io/4.1/api/io/netty/util/concurrent/Future.html), which extends Java's `Future` and has support for listeners.

> I see you removed older versions from the CI file (but not sure if that's intended to stay this way when releasing).

I'm actually working on that now. I THINK it should be possible to get everything working with Java 6 (provided that end users are comfortable using native SSL). I'm actually working on that now, and should have a more informed perspective soon.

jchambers:
@flozano Switched to Netty futures in fdaa85c. You can now add listeners to your heart's content ;)

Good to hear from you, by the way!

flozano:
That's great! 

yeah, good to be around here again :dancers: and good to see the death of low-level TCP-based APNS is near !

I'm totally OK with exposing the netty guts of pushy, so for me it's not a problem to have netty futures there. After this commit, a wrapper to expose CompletableFuture / guava ListenableFuture / Spring ListenableFuture / whatever is trivial and doesn't require any polling, so that's perfect!

buyukim:
I would like to say thank you for trying to get this to work with Java 6.  We are stuck with an older version of IBM WebSphere and will not be upgrading until WAS 9 / Java 8 will be released later this year / next year.

jchambers:
> I would like to say thank you for trying to get this to work with Java 6.

You're welcome!

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -0,0 +1,749 @@
...
+ * number of threads in check.</p>
+ *
+ * <p>Once a client has been constructed, it must connect to an APNs server before it can begin sending push
+ * notifications. Apple provides a production and development gateway; see {@link ApnsClient#PRODUCTION_APNS_HOST} and
+ * {@link ApnsClient#DEVELOPMENT_APNS_HOST}. See the
+ * <a href="https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/APNsProviderAPI.html#//apple_ref/doc/uid/TP40008194-CH101-SW1">APNs
+ * Provider API</a> documentation for additional details.</p>
+ *
+ * <p>Once a connection has been established, an APNs client will attempt to restore that connection automatically if
+ * the connection closes unexpectedly. APNs clients employee an exponential back-off strategy to manage the rate of
```
jmason: s/employee/employ/

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -0,0 +1,749 @@
...
+                            ApnsClient.this.connectionReadyPromise.tryFailure(
+                                    new IllegalStateException("Channel closed before HTTP/2 preface completed."));
+
+                            synchronized (ApnsClient.this.bootstrap) {
+                                ApnsClient.this.connectionReadyPromise = null;
+
+                                if (ApnsClient.this.reconnectionPromise != null) {
+                                    log.debug("Disconnected. Next automatic reconnection attempt in {} seconds.", ApnsClient.this.reconnectDelay);
+
+                                    future.channel().eventLoop().schedule(new Runnable() {
```
jmason: One thing that we have found useful in production is limiting the initial number of connection attempts (to 10) with a max reconnectDelay of 60 seconds between them, using jittered exponential backoff.  (A limited number of attempts means that a bad SSL cert won't retry indefinitely, always failing to connect.)

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -0,0 +1,749 @@
...
+                            if (future.isSuccess()) {
+                                synchronized (ApnsClient.this.bootstrap) {
+                                    if (ApnsClient.this.reconnectionPromise != null) {
+                                        log.info("Connection to {} restored.", future.channel().remoteAddress());
+                                        ApnsClient.this.reconnectionPromise.trySuccess();
+                                    } else {
+                                        log.info("Connected to {}.", future.channel().remoteAddress());
+                                    }
+
+                                    ApnsClient.this.reconnectDelay = INITIAL_RECONNECT_DELAY;
```
jmason: And another useful feature we've found is to switch to a different max reconnectDelay (and max retries, see above) once the initial connect is successfully established -- ie. we have now determined that the SSL cert works ok.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -0,0 +1,749 @@
...
+                            // We always want to try to fail the "connection ready" promise if the connection closes; if
+                            // it has already succeeded, this will have no effect.
+                            ApnsClient.this.connectionReadyPromise.tryFailure(
+                                    new IllegalStateException("Channel closed before HTTP/2 preface completed."));
+
+                            synchronized (ApnsClient.this.bootstrap) {
+                                ApnsClient.this.connectionReadyPromise = null;
+
+                                if (ApnsClient.this.reconnectionPromise != null) {
+                                    log.debug("Disconnected. Next automatic reconnection attempt in {} seconds.", ApnsClient.this.reconnectDelay);
```
jmason: it'd be really nice at this point (and other similar spots) to provide metric hooks, so we can get a callback to increment our internal service metrics, for monitoring.  (I'd be happy to send in a followup pull request at some point, if that sounds useful?)

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -0,0 +1,749 @@
...
+     * @param notification the notification to send to the APNs gateway
+     *
+     * @return a {@code Future} that will complete when the notification has been either accepted or rejected by the
+     * APNs gateway
+     */
+    public Future<PushNotificationResponse<T>> sendNotification(final T notification) {
+
+        final Future<PushNotificationResponse<T>> responseFuture;
+
+        // Instead of synchronizing here, we keep a final reference to the connection ready promise. We can get away
```
jmason: hmm, I'm not 100% sure about this.

I would suggest using a volatile boolean, initialized to false at class creation, and if that's still false then enter a synchronized block, setting up the promise and setting the boolean to true.   That way, the fast path is just a volatile access, but it's definitely threadsafe.  (This is how I worked around https://github.com/relayrides/pushy/issues/189 in our calling code, fwiw)

jmason:
This is looking great ;)  Really looking forward to trying it out (and deleting large chunks of _our_ code too)

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: Oops. Good catch. Fixed in 7714f92.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: So my thinking here—and I'm very open to other ideas on this—is that most of the permanent failures (i.e. bogus certificates) would happen on the initial connection attempt. We do NOT automatically retry the initial connection attempt (automatic reconnection attempts only start after we establish the initial connection), and the idea is that we want callers to see the failure and deal with it accordingly. After that, it's more likely that we're dealing with a temporary network outage or a server shutdown or something.

I guess I see this as a choice between two evils:

1. We decline to impose a retry limit on automatic reconnection attempts and (gently!) spin forever if a certificate expires after the initial connection attempt.
2. We *do* impose a retry limit, but guess wrong and wind up permanently disconnecting a client that would have automatically reconnected later.

What do you think?

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: We won't start initial reconnection attempts at all until we've established the initial connection. Does that seem reasonable to you?

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: Definitely. This is something we've wanted to do for a while (see #175), and I think the switch to HTTP/2 leaves us in a good position to actually get it done. Now that we can actually do things like measure throughput and outcomes and healthy connections, it's much easier to imagine what we should actually instrument.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: It could, but we still run the risk of a `volatile` promise getting set to `null` by another thread halfway through this method (i.e. the first read will be non-null, but subsequent reads will be null and cause exceptions).

The local reference approach means that the state of the promise might change (i.e. the connection might close), but at least it won't mysteriously go `null` on us.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jmason: hmm.  I agree that the permanent failures should show up on the first attempt, but I *do* think there's value in retrying that initial connection attempt, all the same.

I think there are a subset of failure scenarios where that may fail, even if the cert is fine -- one would be DNS resolver slowness, where the DNS name may have been resolved by the time retry number 2 rolls around for instance, or even just a transient network glitch which caused the first connection attempt to fail/timeout.

Is that possibly under caller control?  ie if they see the initial failure, can they apply their own retrying logic and re-call connect() as they see fit?

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jmason: excellent! ;)

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jmason: oh yeah -- I mean in addition to the local reference, ie. the local ref is still a local copy of the (now-volatile) member.

The risk I see is that one thread calling connect() will cause the connectionReadyPromise member to be created, but another thread calling sendNotification() may still see the null value. Marking it volatile avoids that risk.

jchambers: Ah—I see where you're going. Made both `connectionReadyPromise` and `reconnectionPromise` `volatile` in d8ff986.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: > Is that possibly under caller control? ie if they see the initial failure, can they apply their own retrying logic and re-call connect() as they see fit?

Yes! Callers could certainly do something like this:

```java

boolean connected = false;

while (!connected) {
    final Future<Void> connectionFuture = apnsClient.connect(host);
    connectionFuture.await();

    connected = connectionFuture.isSuccess();
}
```

> I do think there's value in retrying that initial connection attempt, all the same.

I can see that, but given that there's already a mechanism for callers to manage this kind of thing themselves, I'm inclined to view limited automatic retries on initial connection attempts (whew) as an enhancement for a follow-up release. Seem legit?

jmason: Yep, that sounds perfect!

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jmason: as discussed above, this is a non-issue given the ability to control initial-connect retries externally.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jmason: great, thanks.

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
stela: Regarding SSL certificates, you can do a check against expired ones by iterating through a keystore's aliases' certificate chains. For each X509Certificate in a chain, assure that getNotAfter() is in the future. If not, print out a warning indicating which certificate is expired (get the subject through getSubjectX500Principal()) and you might as well give up (re)connecting. It's also a good idea to continuously monitor and warn about certificates about to expire in the near future.

jchambers: That's definitely true, but it seems like there's some badness either way. Whether or not we stop the automatic reconnection attempts, callers are left in a situation where their client has stopped and will not reconnect. As near as I can tell, most of what we'd gain by stopping reconnection attempts after certificate expiration is a modest reduction in resource usage (which should already be very low thanks to exponential back-off) until somebody intervenes.

It's true that we could offer users a clear warning about expiration if we think it might be a problem, but I suspect that will also be clear from the exceptions generated when reconnection attempts fail. Given that we could also conceivably get things wrong with this kind of check (and accidentally clobber clients that would otherwise reconnect), I'm inclined to leave certificate expiration handling out of Pushy itself and trust other layers of the whole system to deal with it.

Seem okay?

PR #161, opened 2015-03-29
Add socks proxy support without breaking netty 4.0.x compatibility.

Nowheresly:
Hi Jon,

here is a PR for the issue #99.
I use your great library to send notifications, but since I'm working for a large organization with strict security rules, I absolutely need to use a proxy.
My notification batch runs inside an internal network that is NOT allowed to directly reaches the internet. All connections to the outside world need to go through a proxy.
As you said previously, netty currently does NOT support proxy as of it's version 4.0. But since I can't wait until netty 4.1 is out, I tried to upgrade to netty-all-4.1.0.Beta4 and see what needs to be done.
In fact, it was really simple:
- first, I upgraded to netty-all-4.1.0.Beta4 and good news, pushy still works like a charm!
- next, in the pipeline, I added the socks5 proxy handler at the very first position;
- finally, since the domain name apple.com is not reachable from inside our internal network, I added a NameResolver strategy to tell netty not to resolve this specific domain name.
And it works!!!

But to do so, I had no choice but to redefine the ApnsConnection class with my custom socks5 configuration.
That's the reason why I submit this PR. The goal is to keep the compatibility to netty 4.0.x and providing some hooks so that pushy can be configured to be used behind a proxy with netty 4.1.

Please note this PR contains an update of the documentation to explain how to configure proxy settings.

Could you spend some time reviewing this PR and give me your feedback?
Thanks!

PR #215, opened 2016-01-19
Proxy support (closes #99)

### LINKED ISSUE #99: configure pushy to use a proxy
It might be useful for other people (it's useful for me) to add a way to configure a proxy when sending push notifications 

Right now I'm using javapns and they provide a way to configure the proxy, but I can't figure out how to do it with pushy or even if it's possible.

jchambers:
This adds a `ProxyHandlerFactory` that callers may use to inject proxy handlers into an `ApnsClient`'s pipeline. This is very much inspired by #161, with thanks to @Nowheresly.


jchambers:
This is looking good as near as I can tell from "laboratory" testing. Going to go ahead and merge it.

PR #216, opened 2016-01-19
Remove explicit hpack dependency

panchenko:
It was added together with netty-all which is supposed to be used
when managing dependencies manually, so is not needed after
switching to netty-codec-http2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,12 +34,6 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.6</version>
         </dependency>
-        <!-- TODO This should go away; needing it here at all appears to be a Netty bug. -->
-        <dependency>
-            <groupId>com.twitter</groupId>
-            <artifactId>hpack</artifactId>
-            <version>0.11.0</version>
-        </dependency>
         <!-- Test dependencies -->
         <dependency>
             <groupId>junit</groupId>


jchambers:
@panchenko Can you elaborate on why this isn't needed any more? My (probably incorrect!) understanding was that this was something of a goof where Netty internalized the HPACK stuff, but accidentally kept references to the old Twitter HPACK classes. As near as I can tell, it's a non-issue in Netty 4.1 CR1, but is still an issue in 4.1 Beta 8. Was I just misunderstanding things before?

panchenko:
@jchambers  it was added in the same commit as netty-all, which has netty-xxx dependencies as optonal in [pom](http://central.maven.org/maven2/io/netty/netty-all/4.1.0.Beta8/netty-all-4.1.0.Beta8.pom). 

Usually projects build "all" artifact for those who don't want to use a proper (maven) dependency management and want to download a single file instead. It contains code from all the subprojects but without any dependencies.

After that the project manifest was changed to use only netty-codec-http2 which correctly has this dependency in [pom](http://central.maven.org/maven2/io/netty/netty-codec-http2/4.1.0.Beta8/netty-codec-http2-4.1.0.Beta8.pom), so it would be better to rely on that and it is not necessary to repeat hpack dependency.

panchenko:
hpack version is specified in [netty-parent pom](http://central.maven.org/maven2/io/netty/netty-parent/4.1.0.Beta8/netty-parent-4.1.0.Beta8.pom) in the case of netty-code-http2

jchambers:
Makes sense to me. Thanks for the explanation, and thanks for the contribution!

PR #217, opened 2016-01-19
Fix passing not specified password for .p12 file

panchenko:
KeyStore.PasswordProtection constructor accepts null,
but the next KeyStore.Builder.newInstance() call doesn't accept protection as null

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -245,7 +245,7 @@ public class ApnsClient<T extends ApnsPushNotification> {
         final PrivateKey privateKey;
 
         try {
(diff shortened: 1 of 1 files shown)


jchambers:
Oops. Good catch. Will merge this as soon as I have some unit tests ready to go.

jchambers:
@panchenko As a matter of curiosity, how are you generating PKCS#12 files without password protection?

jchambers:
To elaborate:

I agree that this change prevents `NullPointerExceptions`, which is undeniably a good thing! At the same time, I've never managed to generate a PKCS#12 file with no passphrase at all. I've managed to generate one with an *empty, but still present* passphrase like this:

```sh
openssl pkcs12 -export -in single-topic-client.pem -inkey single-topic-client.key -out single-topic-client-unprotected.p12 -nodes -password pass:
```

…but then the password is an empty string (`""`) rather than `null`. Is that not true for you, too?

PR #222, opened 2016-01-25
Add metrics interfaces and members, to demo a way to offer metrics wihout tying down to a specific implementation

jmason:
I've spiked up a quick demo of what I would say is a reasonable way to offer metrics support in Pushy, without tying that down to a specific backend implementation too tightly.

This should allow implementation against:

* Dropwizard Metrics (v3),
* Yammer Metrics (v2),
* Datadog: https://github.com/indeedeng/java-dogstatsd-client
* and statsd using an interface similar to https://github.com/tim-group/java-statsd-client .

Essentially, the API used by Pushy internally is a simplified version of the Metrics API, but should be implementable for statsd and Datadog easily enough too.  I'd be curious if there are other metrics systems this approach would pose a problem for....

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -34,6 +34,7 @@ import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLException;
 
+import com.relayrides.pushy.apns.metrics.*;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


jchambers:
@jmason Thanks a million for this!

For the audience at home, this led to a great discussion in #175. We decided to pursue a slightly different approach (see #251), but that approach is very much inspired by this demo.

I'm going to close this, since I think it's served its purpose admirably. Thanks again!

jmason:
cheers Jon ;)

PR #248, opened 2016-03-05
Introducing jetty-alpn-agent for more clear pom file

matzew:
Removes the pain of adding more and more bogus profiles to the pom

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,26 +39,20 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.6</version>
         </dependency>
+        <dependency>
+          <groupId>org.eclipse.jetty.alpn</groupId>
+          <artifactId>alpn-api</artifactId>
+          <version>1.1.2.v20150522</version>
+          <!-- Provided by alpn-boot; see
+          http://www.eclipse.org/jetty/documentation/current/alpn-chapter.html#alpn-understanding -->
+          <scope>provided</scope>
+        </dependency>
         <!-- Test dependencies -->
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.11</version>
             <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.mortbay.jetty.alpn</groupId>
-            <artifactId>alpn-boot</artifactId>
-            <version>${jetty.alpn.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty.alpn</groupId>
-            <artifactId>alpn-api</artifactId>
-            <version>1.1.2.v20150522</version>
-            <!-- Provided by alpn-boot; see
-            http://www.eclipse.org/jetty/documentation/9.2.8.v20150217/alpn-chapter.html#alpn-understanding -->
-            <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


jchambers:
Awesome. I think we might need to update the README, too, though. We can probably merge this as-is, though, and I can just chase it with a quick update.

pom.xml
```diff
@@ -82,8 +74,11 @@
     <properties>
         <netty.version>4.1.0.CR3</netty.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <jetty.alpn.path>${settings.localRepository}/org/mortbay/jetty/alpn/alpn-boot/${jetty.alpn.version}/alpn-boot-${jetty.alpn.version}.jar</jetty.alpn.path>
-        <argLine.bootcp>-Xbootclasspath/p:${jetty.alpn.path}</argLine.bootcp>
+        <alpn-api.version>1.1.2.v20150522</alpn-api.version>
```
jchambers: I get why we want to define version numbers for things like `alpAgent` in properties, since we use them in several places in here, and we definitely want to be consistent. Would it make more sense to inline `alpn-api.version`, though? It looks like we only use it in one place.

I'm very open to arguments that, since it's sensitive to the version of Netty we're using, the ALPN API version and Netty version should be close to each other for readability.

matzew: @jchambers ok, no longr using a variable for `alpn-api.version`

matzew:
@jchambers yes, the README should be touched too :) 

Mind taking a shot at it ?

pom.xml
```diff
@@ -47,17 +47,9 @@
...
-            <version>${jetty.alpn.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.eclipse.jetty.alpn</groupId>
             <artifactId>alpn-api</artifactId>
             <version>1.1.2.v20150522</version>
-            <!-- Provided by alpn-boot; see
-            http://www.eclipse.org/jetty/documentation/9.2.8.v20150217/alpn-chapter.html#alpn-understanding -->
             <scope>provided</scope>
```
jchambers: One more sanity check: is this still provided by the ALPN agent, or should we change the scope?

matzew: removed it - not needed; see also https://github.com/jetty-project/jetty-alpn-agent

jchambers: Forgive me if I'm being dense; I was just trying to write an update for the README that explains *why* it's not needed, but I'm not quite finding what I think you're referring to on the `jetty-alpn-agent` project page. I hate to be needy, but can you be a little more specific about where I should be looking?

Thanks!

matzew: I just checked netty (they use the agent, instead of all the profiles too). Looks like the `alpn-api` is still there, as provided:
https://github.com/netty/netty/blob/c295f2239049a677527fa2be7ab14d7e413da4b9/pom.xml#L291

I _guess_ best is, to follow that...
I

matzew:
@jchambers I also added a bit of doc around the agent.

let me know if more is needed before merging this

matzew:
@jchambers wanna merge it? Or do you need more/different `README` update?

jchambers:
I just need six adjacent brain cells to devote to reading things ;)

Should be able to give this attention it deserves shortly. Thanks for your patience!

jchambers:
@matzew Looks good to me! Thanks for the contribution and for your patience.

PR #251, opened 2016-03-12
Metrics! (closes #175)

### LINKED ISSUE #175: Expose important metrics

jchambers:
This adds an interface callers can use to listen for performance metrics from `ApnsClients`. The goal here is to expose important metrics without introducing dependencies on specific metrics libraries.

TODO:

- [x] Update the README (04275a7)
- [x] ~~Decide where to put reference implementations~~ (deferred)


src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -533,9 +541,15 @@ public void setProxyHandlerFactory(final ProxyHandlerFactory proxyHandlerFactory
                     new IllegalStateException("Client's event loop group has been shut down and cannot be restarted."));
         } else {
             synchronized (this.bootstrap) {
+                final ApnsClientMetricsListener metricsListener = this.metricsListener;
+
                 // We only want to begin a connection attempt if one is not already in progress or complete; if we already
                 // have a connection future, just return the existing promise.
                 if (this.connectionReadyPromise == null) {
+                    if (metricsListener != null) {
```
jmason: I would be inclined to create a NullMetricsListener class which just contains no-ops, and use that by default, in order to avoid the need to null-check this on each call (as in NullMetrics in https://github.com/relayrides/pushy/pull/222/files#diff-bd45fb4e15d6ed613f0e6f1e1d5a193fR6 ).

jmason:
looks great!

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: Fair enough. Done in 7079143.

jchambers:
I think I'm going to punt on a reference implementation for now, at least in part because I don't want to hold up the v0.6 release while I figure out how to do a multi-module release with Maven. I'm also thinking it could make sense to do vendor-specific implementations as entirely separate projects. I'm open to suggestions here, though!

jmason:
:tada: looks awesome. looking forward to using this ;)

PR #255, opened 2016-03-22
Added support for all kinds of netty ELGs.

bfg:
This patch adds ability to use any kind of event loop
group with pushy.

Tested with epoll on linux.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -39,6 +39,10 @@ import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLException;
 
+import io.netty.channel.Channel;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.oio.OioEventLoopGroup;
+import io.netty.channel.socket.oio.OioSocketChannel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


jchambers:
I like the idea, but think we need to figure out what's going on with those failing tests (confession: I have not made even the most feeble of attempts to investigate).

bfg:
Hm, i wondered what's up with the tests too; All of them failed when i tried to build master and 0.5.2, tha't why i didn't bother with them.

Now i investigated a bit and it looks that all tests always fail with openjdk (i have ```openjdk version "1.8.0_01-internal``` installed on my linux x86_64 laptop), but they do work with oracle jdk ```1.8.0_74-b02```.

Then i investigated why this is happening and it's because of ```ApnsClientTest``` initializes instance with null event loop group parameter and my patch threw an exception in such cases. I've updated the patch so that is not so strict and returns NioSocketChannel class in case when null event loop group is passed to ```getSocketChannelClass()```

src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -443,6 +449,45 @@ protected void handshakeFailure(final ChannelHandlerContext context, final Throw
...
+
+        if (eventLoopGroup instanceof NioEventLoopGroup) {
+            return NioSocketChannel.class;
+        } else if (eventLoopGroup instanceof OioEventLoopGroup) {
+            return OioSocketChannel.class;
+        }
+
+        // epoll?
+        String className = eventLoopGroup.getClass().getName();
+        if (EPOLL_EVENT_LOOP_GROUP_CLASS.equals(className)) {
```
jchambers: To make sure I understand: are we not doing `instanceof` here because the epoll classes live in a separate library, and may not be on the classpath? If that's the case, would it make sense to use the same approach (matching class names) the whole way through for consistency? We could have a map of event loop group class names to socket class names, or could use a `switch` block. At the very least, could we add a comment explaining why we need to handle epoll differently?

bfg: Only EpollEventLoopGroup lives in a separate java artifact (netty-native-epoll-linux-<arch>). We're doing instanceof here because it's cheap to code and because NioXXX and OioXXX classes are always available.

I've implemented something similar also for netty, see https://github.com/netty/netty/pull/5050, because pushy is like third library that needs this kind of patch in order to support epoll on linux.

jchambers:
:metal:

Thanks, @bfg!

jchambers:
Also, @bfg, why did you remove those `@SuppressWarning` annotations?

bfg:
Damn it, it was my intellij!!!

jchambers:
> Damn it, it was my intellij!!!

Ha—all good. Thanks for clarifying ;)

Issue #264, opened 2016-04-05
interrupted persistent connection


pedobry:
I use pushy 0.6 which establish persistent connection. However, my firewall drops idle TCP connection after 40 minutes and unfortunately the bad way (no RST, just silent drop). When I try to send notification to that connection, I get "Connection time out" exception, but only after 15 minutes (that's default Linux OS timeout to retransmit the packet). And that's not very convenient.

I'l change Linux settings /proc/sys/net/ipv4/tcp_retries2 and /proc/sys/net/ipv4/tcp_retries1 to some smaller values, but it would be really helpful if pushy can enable ChannelOption SO_KEEPALIVE on the socket.

jchambers:
I think we have this covered in #266.

Issue #257, opened 2016-03-28
Connection reset by peer


abraxascorner:
Hi,

I am using client for sending both standard and voip push messages and have the following exception occurs from time to time in my logs:
```
WARN - ApnsClientHandler - APNs client pipeline caught an exception.
java.io.IOException: Connection reset by peer
	at sun.nio.ch.FileDispatcherImpl.read0(Native Method)
	at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:39)
	at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
	at sun.nio.ch.IOUtil.read(IOUtil.java:192)
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:380)
	at io.netty.buffer.PooledUnsafeDirectByteBuf.setBytes(PooledUnsafeDirectByteBuf.java:221)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:871)
	at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:245)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:112)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:510)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:467)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:381)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:353)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:742)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137)
```
I am connecting client on the load of my app and some other application calls mine when I need to send voip notification:
client instantiation:
```java
@PostConstruct
private void init() {
	try {
		client = new ApnsClient<>(new File(keyStoreFilePath), keyStorePassword);

		client.connect(ApnsClient.DEVELOPMENT_APNS_HOST).addListener(future -> logger.info("APNS client initialized and connected to gateway."));
	}
	catch (SSLException e) {
		logger.error("Failed to initialize APNS client. {}", e.getMessage());
	}
}
```
[...]

PR #265, opened 2016-04-06
allow to enable TCP keepalive

pedobry:
Allow to enable TCP keepalive on socket, so OS can keep the persistent connection alive and avoid idle connection on firewalls.

### DIFF
--- a/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -520,6 +520,19 @@ public class ApnsClient<T extends ApnsPushNotification> {
     }
 
     /**
+     * Enable TCP keepalive packets for the connection. Keepalive timeouts are managed by underlying OS.
+     * For Linux is default keepalive timeout 7200 seconds.
+     *
+     * @param enabled enable sending periodic TCP keepalive packets
+     *
+     */
+    public void setKeepalive(final boolean enabled) {
+        synchronized (this.bootstrap) {
+            this.bootstrap.option(ChannelOption.SO_KEEPALIVE, enabled);
+        }
+    }
+
+    /**
      * Sets the metrics listener for this client.
      *
      * @param metricsListener the metrics listener for this client, or {@code null} if this client should not report


jchambers:
I think this is a problem worth solving, but I have two quibbles:

1. Why not just turn on `SO_KEEPALIVE` all the time?
2. You pointed out one TCP-level connection closure problem in #264, but @abraxascorner pointed out a related-but-slightly-different problem in #257: connections will get closed by the APNs server if they sit idle for too long. I don't think `SO_KEEPALIVE` will save us there.

What I'd actually like to do is send HTTP/2 `PING` frames if the connection is idle for a while (a minute?); I think that should solve both problems, no? In addition, it will help us make sure the connection stays healthy at an application level, and it seems like that might be the better approach. I don't think there's any material harm in using both approaches, but if we take the `PING` approach, `SO_KEEPALIVE` would basically be vestigial, right?

jmason:
I would be in favour of HTTP-level PINGs rather than relying on TCP Keep-Alive -- the latter is only manipulable on a system-wide level in Linux, afaik, and defaults to 2 hours, which is probably far longer than most problematic firewalls' timeouts.

pedobry:
Technically, there is no harm in enabling `SO_KEEPALIVE` permanently. I just like the possibility to disable/enable it if I need to. 
I agree that default 2 hours TCP keepalive timeout on Linux is unfortunate and enabling `SO_KEEPALIVE` alone won't help to keep connection open through most of the firewalls. Adjusting `/proc/sys/net/ipv4/tcp_keepalive_time` is necessary too.

Solution with PING would be far more elegant and simpler.
Issue #257 might or might not be related to Apple servers sending GOAWAY. As they might send unsolicited GOAWAY anytime for any of those reasons listed in API, it should be somehow handled too.

jchambers:
> As they might send unsolicited GOAWAY anytime for any of those reasons listed in API, it should be somehow handled too.

I'm preeeeeeeetty sure we do that even if we don't report it externally, but now I'm starting to second-guess things. I'll double-check on that.

jchambers:
> I'll double-check on that.

I double-checked, and this is all working as expected.

I went ahead and did the `PING` thing in #266. @pedobry with sincere thanks for the suggestion, I'm going to close this for now, since I think sending pings will be a more effective solution for more users. Sound good?

pedobry:
Yes. Using HTTP/2 PING is definitely better solution. Thanks for implementing this so quickly.

PR #267, opened 2016-04-10
Optimize payload construction (closes #11)

### LINKED ISSUE #11: Optimize APNs payload construction
To make sure an APNs payload fits within a certain size, we repeatedly regenerate the payload JSON object with different message body lengths. Our current approach could be improved in two important ways:

1. Just regenerate the message body string instead of the entire payload
2. Use a binary search instead of a linear walk to seek to the right message body length (maybe?)

jchambers:
#11 has been on the to-do list for a really long time, and this finally clears it out. This does two important things:

1. It avoids frequently modifying the payload map and regenerating the entire notification payload.
2. It uses a binary search instead of a linear search to identify maximal-length alert bodies. This may actually turn out to be very slightly slower for users who were sending mostly-ASCII messages, but will be an enormous improvement for anybody sending mostly-Unicode messages.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.5</version>
+            <version>2.6.2</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -32,7 +32,7 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.5</version>
+            <version>2.6.2</version>
```
jchambers: We need to update Gson since 2.6 is the first series that allows a `String` to be a top-level Json text.

PR #271, opened 2016-04-14
Latest of netty

matzew:
Updated the version of Netty to 4.1.0.CR7, to keep up with the latest candidate release of Netty.


jchambers:
Thanks as always!

PR #284, opened 2016-05-03
Maven modules

jchambers:
Let me begin this pull request by pointing out that I have no practical experience with Maven modules, and any feedback here would be greatly appreciated. That said, the idea here is to break the project into modules so we can do things like add [microbenchmarks](http://openjdk.java.net/projects/code-tools/jmh/) and have a home for things like metrics libraries. Right now, there's just one module (the main project).


pom.xml
```diff
@@ -1,76 +1,37 @@
...
     <description>A Java library for sending push notifications</description>
+
     <parent>
         <groupId>org.sonatype.oss</groupId>
         <artifactId>oss-parent</artifactId>
         <version>7</version>
     </parent>
+
+    <modules>
+        <module>pushy</module>
```
matzew: perhaps name it `pushy-core`, while others could be `pushy-metrics` etc ?

pushy/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    <artifactId>pushy</artifactId>
+    <name>Pushy</name>
+
+    <parent>
+        <groupId>com.relayrides</groupId>
+        <artifactId>pushy-parent</artifactId>
+        <version>0.7.2-SNAPSHOT</version>
+    </parent>
+
+    <dependencies>
```
matzew: It might be that some are being used on the other moduels, in that case I'd move them up, again .

jchambers: Agreed. I'm inclined to deal with that when it actually happens, though. Does that seem reasonable to you?

matzew: yes, why 'expose' all to all submodules :)

pom.xml
jchambers: Fair enough. Done in 847954e.

matzew:
👍

matzew:
one thing. when the artifact is now named `-core` - perhaps version bump to 0.8.0 ?

jchambers:
> when the artifact is now named `-core` - perhaps version bump to 0.8.0 ?

This could be me misunderstanding, but I don't think changing the module name necessarily changes the `artifactId`, right? As far as I can tell, the `pushy-core` *module* still produces an *artifact* called `pushy`.

pom.xml
jchambers: Actually, I think I changed my mind here if only for the petty reason that having everything start with `pushy-` was making autocomplete slightly annoying ;)

Besides—we can always change it later.

matzew:
sounds good

On Sun, May 8, 2016 at 5:51 PM, Jon Chambers <notifications@github.com>
wrote:

> Merged #284 <https://github.com/relayrides/pushy/pull/284>.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/relayrides/pushy/pull/284#event-653738261>
>



-- 
Matthias Wessendorf

blog: http://matthiaswessendorf.wordpress.com/
sessions: http://www.slideshare.net/mwessendorf
twitter: http://twitter.com/mwessendorf

Issue #282, opened 2016-04-30
Payload generation optimization makes performance worse


jchambers:
Well, I goofed. I flipped the numbers on my benchmark results for the payload generation optimization introduced in #267, and the new approach is actually slower than the old approach. There are two things to do here:

1. Actually make payload generation faster.
2. Move to a more robust benchmarking strategy.

For the latter, I'll be looking to add [jmh](http://openjdk.java.net/projects/code-tools/jmh/) to the project.

jchambers:
Well, this has certainly turned into quite the comedy of errors. Yes, I did flip the results of the previous benchmarks, but it turns out those benchmarks were bogus. Using the microbenchmarks introduced in #288, we can show that the changes introduced in #267 WERE a significant improvement:

### 0.6.1

| Benchmark                   | Mode  | Cnt | Score        | Error    | Units |
|-----------------------------|-------|-----|--------------|----------|-------|
| testLongAsciiMessageBody    | thrpt | 80  | 7.958 ±      | 0.158    | ops/s |
| testLongChineseMessageBody  | thrpt | 80  | 5.192 ±      | 0.027    | ops/s |
| testShortAsciiMessageBody   | thrpt | 80  | 116843.074 ± | 1220.330 | ops/s |
| testShortChineseMessageBody | thrpt | 80  | 155858.086 ± | 651.504  | ops/s |

### 0.7.2

| Benchmark                   |  Mode | Cnt | Score        | Error   | Units |
|-----------------------------|-------|-----|--------------|---------|-------|
| testLongAsciiMessageBody    | thrpt | 80  | 1189.459 ±   | 42.306  | ops/s |
| testLongChineseMessageBody  | thrpt | 80  | 3175.659 ±   | 13.284  | ops/s |
| testShortAsciiMessageBody   | thrpt | 80  | 119996.542 ± | 888.424 | ops/s |
| testShortChineseMessageBody | thrpt | 80  | 155339.674 ± | 852.837 | ops/s |

So, in short, the new approach is about the same when messages don't have to be trimmed (which makes sense) and three orders of magnitude faster in some cases when they DO need to be trimmed. In the process of building and running the benchmarks, we DID discover (and fix!) a separate issue that could cause payload generation to hang. See #287 for details.

I think *this* issue is now under control, and will put out a point release to address the hang as soon as possible.

PR #288, opened 2016-05-04
Microbenchmarks with jmh

jchambers:
This adds benchmarks using [jmh](http://openjdk.java.net/projects/code-tools/jmh/). It depends on #284, and is primarily motivated to help understand the performance implications of recent payload construction optimizations (see #282).


pushy-benchmark/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilderBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+    private String shortAsciiMessageBody;
+    private String longAsciiMessageBody;
+    private String shortChineseMessageBody;
+    private String longChineseMessageBody;
+
+    @Setup
+    public void setUp() {
+        this.apnsPayloadBuilder = new ApnsPayloadBuilder();
+
+        this.shortAsciiMessageBody = RandomStringUtils.randomAscii(MAXIMUM_PAYLOAD_SIZE / 8);
```
jmason: do you see any differences between re-runs of the test harness?  It might be good to mutate these input messages somehow between iterations, if so, to "smooth out" any effects of the random input data.

pushy-benchmark/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilderBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+            for (int i = 0; i < cjkUnifiedIdeographs.size(); i++) {
+                chineseCharacters[i] = cjkUnifiedIdeographs.get(i);
+            }
+        }
+
+        this.shortChineseMessageBody = RandomStringUtils.random(MAXIMUM_PAYLOAD_SIZE / 8, chineseCharacters);
+        this.longChineseMessageBody = RandomStringUtils.random(MAXIMUM_PAYLOAD_SIZE * 2, chineseCharacters);
+    }
+
+    @Benchmark
```
jmason: does JMH do warmup iterations by default?

pushy-benchmark/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilderBenchmark.java
jchambers: > do you see any differences between re-runs of the test harness?

Actually, no. I would expect more variation if we were using heterogeneous mixes of characters, but everything in the ASCII set takes up one byte in the escaped string, whereas everything in the Chinese character set expands to take up three(?) bytes.

If we had messages that contained a mix of characters with different sizes, we'd definitely want to mutate between runs.

pushy-benchmark/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilderBenchmark.java
jchambers: Yes; the default configuration is 10 runs (forks) with 20 warmup iterations and 20 measurement iterations per benchmark.

jmason: cool

PR #307, opened 2016-06-05
Allow undefined behavior with multi-key P12 files (closes #292)

### LINKED ISSUE #292: P12Util didn't reject a keystore with multiple private keys
As [discussed on the mailing list](https://groups.google.com/d/msg/pushy-apns/3HonwEuUxYk/oBioFgMyFAAJ), it sounds like `P12Util` may not be rejecting all keystores with multiple private key entries.

jchambers:
As discussed in #292, we can't reliably detect problematic P12 files with multiple private keys. Rather than advertise that we're checking for problematic cases (since we know we can't catch the case that users are most likely to encounter), we could just warn users that the behavior with a multi-key P12 file is undefined.

None of the options feel great to me, but I did want to open this for discussion. What do you all think?


jchambers:
Sounds like there aren't any major objections here, so I'm going to go ahead and merge this. We'll keep an eye out for longer-term solutions in the future.

PR #310, opened 2016-06-07
Builder pattern for ApnsClients

jchambers:
The number of options for `ApnsClient` instances was growing, and will probably continue to grow in the future. Many of them also only really made sense to set before starting a client, and making them all constructor arguments was going to get crazy really fast. The idea here is to introduce a builder pattern for `ApnsClient` instances. That should clarify which options should be set at client construction time (basically all of them) without resulting in huge constructors.

TODO:

- [x] Finish documenting `ApnsClientBuilder`
- [x] Update README (75f1ced)
- [x] Figure out certificate chain stuff for `setTrustedServerCertificate`

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
@@ -24,9 +24,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.security.PrivateKey;
 import java.security.KeyStore.PrivateKeyEntry;
 import java.security.KeyStoreException;
+import java.security.PrivateKey;
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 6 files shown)


pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+     */
+    public ApnsClientBuilder<T> setClientCredentials(final X509Certificate clientCertificate, final PrivateKey privateKey, final String privateKeyPassword) {
+        this.clientCertificate = clientCertificate;
+        this.privateKey = privateKey;
+        this.privateKeyPassword = privateKeyPassword;
+
+        return this;
+    }
+
+    protected ApnsClientBuilder<T> setTrustedServerCertificate(final File trustedServerCertificatePemFile) {
```
jchambers: I'd like to publicize this for certificate pinning, but don't yet understand Netty's trust manager construction well enough to be confident in the approach. I intend to do some more research here.

PR #311, opened 2016-06-08
No-op name resolvers for proxies (fixes #309)

### LINKED ISSUE #309: Pushy Proxy Support requires external DNS lookup
Hello, 

we want to switch to Pushy library to be able to use the new APNS Provider API. Until now, we have been using JavaPNS library to send APNS push messages.
As our server is on an internal network, we need the proxy support that became available in Pushy 0.6. (We also talk to GCM for Android Pushes, and all  external communication goes through our proxy.)

I managed to have Pushy (current version 0.7.2) run successfully on my local development environment, but when running on the internal server, I was getting a ClientNotConnectedException.
(This was after I have resolved all issues with SSL and ALPN.)

I then had a network colleague look at the protocol layer on the host when initiating the SSL Connection over the proxy.
Surprisingly, the Library first tried to resolve the name **api.push.apple.com** which failed on the internal server, as this server cannot resolve external host names.

This seems to be the root cause of the _ClientNotConnectedException_, because when we configured a valid IP in the local /etc/hosts file on the internal Linux Server, the Pushy call succeeded.

Our network colleagues also confirmed that the name resolution should be done by the proxy and not by the internal server.

To hardcode the IP Address of that Apple APNS host in the local hosts file is definitely no option for a stable environment, as we don't know if all apple hosts will keep their IP Adresses forever...

Also, all other communication to external systems over proxy worked so far, so that I think that this is a problem specific to the pushy library or the referenced/included libraries.

Can anybody help or advise how to solve that issue?

Thanks and kind regards,

Wolfgang

jchambers:
This fixes an issue where we'd add a proxy handler to the pipeline, but would try to resolve addresses using un-proxied DNS lookups. This caused two classes of problems:

1. In heavily restricted environments, the DNS lookup would fail entirely. See #309.
2. In less restricted environments, the DNS lookup might succeed, but then connections to an already-resolved IP would be blocked by proxies looking for connections to specific domains. See [the mailing list](https://groups.google.com/d/msg/pushy-apns/iYtB7TSi9lU/FKbVE1wdAQAJ) for discussion of this case.

By letting the proxy deal with name resolution, we should avoid both issues.

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -74,6 +74,8 @@ import io.netty.handler.ssl.SslProvider;
 import io.netty.handler.ssl.SupportedCipherSuiteFilter;
 import io.netty.handler.timeout.IdleStateHandler;
 import io.netty.handler.timeout.WriteTimeoutHandler;
+import io.netty.resolver.DefaultAddressResolverGroup;
+import io.netty.resolver.NoopAddressResolverGroup;
 import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.FailedFuture;
 import io.netty.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


jchambers:
This seems fine in "lab" testing; going to go ahead and merge.

PR #316, opened 2016-06-17
added the alpn-api dependency example for Maven users

mohamedhafez:


mohamedhafez:
ah turns out its not necessary to list this dependency at all if you have one of the netty-tcnative dependencies, I misunderstood

PR #317, opened 2016-06-17
Fix certificate name in testExtractApnsTopicsFromMultiTopicCertificate

KHellkvist:
We´re having problems with pushing with certain certs. 
While debugging I found this typo in a test and it took me a minute or two to get it right!
Now no one else needs to get confused!


jchambers:
Oops. Good catch—thank you!

Issue #314, opened 2016-06-15
Add support for Token Authentication


fabiankoebel:
see WWDC 2016: [https://developer.apple.com/videos/play/wwdc2016/724/](https://developer.apple.com/videos/play/wwdc2016/724/)

jchambers:
Sounds like this should cut down on Certificate Woes™ pretty significantly, and will be especially helpful for users working with multiple topics; my understanding is that, since we're authenticating once per request and not once per connection, a single client will be able to work with multiple signing keys/topics.

Some notes from the WWDC video (since there are some details in audio that aren't really clear from the slides):

- We'll send an auth token with every request
  - It stands to reason, but would like to confirm that this means that a single connection (if "armed" with multiple signing keys) can work with multiple topics
- Uses [JSON Web Tokens](https://jwt.io/) to generate auth tokens
  - Algorithm is `ES256`
  - `iss` is the developer team ID (I'm not 100% sure I know what this means, but I believe it's the ten-character alphanumerical ID)
  - `iat` is the timestamp, in *seconds* since the epoch, at which the token was issued
- Tokens will be sent in an HTTP/2 header like this: `authorization = bearer token.string.here`
- New error codes when using tokens
  - `InvalidProviderToken` is fatal and comes with an `HTTP/403` status code
  - `ExpiredProviderToken` is non-fatal (the token needs to be regenerated) and also comes with an `HTTP/403` status
- Tokens expire after an hour

This is supposed to happen "later this year." As a point of comparison, APNs via HTTP/2 was announced at WWDC 2015 with an expected deadline of "later this summer" for a draft of the updated spec and "later this year" for release to production. In the end, both things happened on December 17, 2015. I recognize that the transition to HTTP/2 was probably a bigger deal than a transition to auth tokens, but I think we should be cautious about betting too heavily on any particular timeline. Looking forward to this, though!

jchambers:
Also, while gathering notes, it sounds like OpenJDK users may need to [jump through some hoops](http://armoredbarista.blogspot.com/2013/10/how-to-use-ecc-with-openjdk.html) to be able to use `SHA256withECDSA`.

jchambers:
I've opened a work-in-progress pull request; see #319 to monitor our progress on token authentication.

jchambers:
Looks like upstream support for token authentication has been [released](https://developer.apple.com/news/?id=09222016a). I'll be on vacation all next week, but will hope to polish off #319 when I get back.

[...]

PR #319, opened 2016-06-19
Token authentication

jchambers:
This is a work-in-progress branch for token authentication (see #314). The goal here is to take what we know now (which isn't a whole lot) and build some foundations so we can move quickly when the Real Thing™ happens "later this year."

So far, this has mostly been a demolition effort, since we know clients won't need to authenticate when connecting to a server.


gabek:
Really looking forward to this!

peter-allday:
Hi, any idea when this will be released? I have just received the "Certificate expiring in 30 days" email, and the idea of replacing it with Tokens that don't expire is very appealing.
Thanks
Peter

jchambers:
As mentioned in #314, I've just returned from vacation and hope to finish this shortly.

peter-allday:
Sorry, don't know how I missed your earlier comment. Hope you enjoyed your
vacation.

Tolsi:
So, when this will happen? :)

vader:
Just to let people know when your certificate expires you can still generate a new certificate. That functionality on Apple has not changed albeit that its not that obvious on their dashboard anymore. I just renewed my certificate after an initial panic that I will have to move to tokens before I was ready. Saying that I can't wait to get away from yearly expiring tokens.

matzew:
any comments on landing this ?

jchambers:
> any comments on landing this ?

Situation is still the same; just need to punch it through as soon as other priorities allow, which I'm hoping will be very soon.

jchambers:
Following discussion in #314, I'm going to close this in favor of #397, which is essentially the same thing we have here except with support for TLS-based client authentication.

PR #320, opened 2016-06-20
support subtitle and mutable-content for ios 10

ScienJus:


jchambers:
Hello!

First and foremost, thank you for the contribution! Before we can merge this, I think there are a few things we need to iron out.

1. Would you mind reopening this against `master` instead of `v0.7.3`?
2. I think we need to figure out what's going wrong with [that failing test](https://travis-ci.org/relayrides/pushy/builds/138881866).
3. Is there somewhere I can read more about the `subtitle` field?

Cheers!

ScienJus:
1, 2: ok, that's my bad.
3: [developer.apple.com][1] has not been updated to the ios 10 version.

[1]: https://developer.apple.com/library/prerelease/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/TheNotificationPayload.html#//apple_ref/doc/uid/TP40008194-CH107-SW2

PR #322, opened 2016-06-21
support subtitle and mutable-content for ios 10

ScienJus:


jchambers:
Porting a comment from #320, you wrote:

> developer.apple.com has not been updated to the ios 10 version.

Totally agreed! I guess I'm wondering how you know this exists, though. Was it announced at WWDC? I didn't notice it in the [What's New in the Apple Push Notification Service](https://developer.apple.com/videos/play/wwdc2016/724/) session. Was it in another one?

ScienJus:
@jchambers from [here][1], but just a little about payload.

[1]: https://developer.apple.com/videos/play/wwdc2016/708/

jchambers:
@ScienJus With apologies for the long delay, I've finally taken a look through the WWDC session you mentioned. I did see the bit in the presentation about the `mutable-content` field, but it wasn't immediately clear to me that `subtitle` was actually a thing that could be included in an APNs payload (i.e. the presentation shows it being set client-side, but not necessarily in the payload itself).

Have you tried the `subtitle` field on your own? I'm afraid I don't have an iOS 10 device handy with which to test.

Thanks again for the contribution, and sorry for the slow response!

jchambers:
I've managed to get my hands on an iOS 10 device to test; the `subtitle` field does, indeed, work as expected. @ScienJus I think this is ready to go, but needs to be rebased onto `master`. If I don't hear from you soon, I'll go ahead and do that on my own.

Thanks for the contribution!

pushy/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -63,8 +65,10 @@
     private static final String SOUND_KEY = "sound";
     private static final String CATEGORY_KEY = "category";
     private static final String CONTENT_AVAILABLE_KEY = "content-available";
+    private static final String MUTABLE_CONTENT_KEY = "mutable-content";
 
     private static final String ALERT_TITLE_KEY = "title";
+    private static final String ALERT_SUBTITLE_KEY = "subtitle";
```
jchambers: I wonder if there's also a `subtitle-loc-key` and `subtitle-loc-args` field that hasn't been documented. I'll try to test that a bit later today, too.

ScienJus:
@jchambers Thanks, i will rebase later!

ScienJus:
@jchambers  

> but needs to be rebased onto master. 

Done!

pushy/src/main/java/com/relayrides/pushy/apns/util/ApnsPayloadBuilder.java
jchambers: Testing reveals that these *do* exist. I'm happy to merge this and add them on my own, though.

jchambers:
@ScienJus I'm going to close this in favor of #368, which builds on your changes and adds docs and support for localized subtitles.

PR #323, opened 2016-06-21
Flexible notification topic

SamuelHFerreira:
There is now a load from certificate for those that doesn't want to inform the topic but just use the  at ApnsPushNotification, 
-Topic now can be null, when it happens, automatically uses the certificate UID.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <groupId>com.relayrides</groupId>
     <artifactId>pushy-parent</artifactId>
     <packaging>pom</packaging>
-    <version>0.9</version>
+    <version>0.8.1</version>
     <name>Pushy parent</name>
     <description>A Java library for sending push notifications</description>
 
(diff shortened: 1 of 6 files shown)


jchambers:
First and foremost, thanks very much for the contribution!

Regrettably, I think we may have to pass on this one. The next big protocol update from Apple introduces authentication tokens (see #314 for additional discussion). One of the big changes there is that we'll be working with keys instead of certificates as our main unit of identity-verification currency. Unlike certificates, keys generally don't have any metadata attached to them, so we probably won't be able to get (for example) a topic list from the signing keys that we'll be working with in the near future.

Since working with keys will almost certainly mean that users will need to be *more* disciplined about topics in the near future, I'd rather not introduce a feature that allows users to be more lax with topics that we'll just have to revert in a couple months.

Does that make sense to you?

Thanks again!

jchambers:
@SamuelHFerreira Again, with sincere thanks for the contribution, I'm going to go ahead and close this for now. Please do post a comment here if you object and we'll revisit.

Cheers!

SamuelHFerreira:
Thank you for the feedback,
i didn't realize that change from Apple, that should be interesting at the next improvement step of this lib,

jchambers:
Awesome. Thank you again!

PR #324, opened 2016-06-22
Mention that Pushy is also compatible with Safari push notifications

mohamedhafez:


jchambers:
Thanks! One minor thing, though: would you mind re-opening this against `master`?

PR #326, opened 2016-06-22
Auto-flush

jchambers:
This moves to using Netty's nascent auto-flush feature instead of flushing explicitly. The main benefit here is simplicity. Depends on ~~https://github.com/netty/netty/pull/5395~~ ~~https://github.com/netty/netty/pull/5445~~ ~~https://github.com/netty/netty/pull/5395~~ https://github.com/netty/netty/pull/5716, which is currently scheduled for Netty ~~4.1.2~~ ~~4.1.3~~ ~~4.1.4~~ oh who even knows any more.

TODO:
- [ ] Wait for Netty ~~4.1.2~~ ~~4.1.3~~ ~~4.1.4~~
- [x] Make sure the benchmarks still look good after this change

### DIFF
--- a/benchmark/src/main/java/com/relayrides/pushy/apns/ApnsClientBenchmark.java
+++ b/benchmark/src/main/java/com/relayrides/pushy/apns/ApnsClientBenchmark.java
@@ -4,8 +4,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-
 import org.apache.commons.lang3.RandomStringUtils;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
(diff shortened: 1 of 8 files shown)


jchambers:
I ran some benchmarks:

## Without auto-flush

```
Benchmark              (flushImmediately)  Mode  Cnt  Score   Error  Units
testSendNotifications                true    ss   40  0.626 ± 0.020   s/op
testSendNotifications               false    ss   40  0.602 ± 0.017   s/op
```

## With auto-flush

```
Benchmark              Mode  Cnt  Score   Error  Units
testSendNotifications    ss   40  0.604 ± 0.026   s/op
```

So, in short, we see similar performance with auto-flush and with the current deferred-flush scheme. Given the gains in simplicity, I think it's definitely a good trade.

jchambers:
Progress on the upstream branch seems to have stalled, so I'm going to remove this from the v0.8 milestone for now and will pursue an alternative solution to #332.

jchambers:
Closing this for now; upstream support for auto-flush appears to have stalled. Will reopen when there's some movement on Netty's side.

PR #349, opened 2016-07-27
Make `ApnsClient#sendNotification` generic (closes #342)

### LINKED ISSUE #342: Question: Why does `ApnsClient` have to be generic?
Maybe I'm misunderstanding `ApnsClient`. I've only coded in it for maybe 30 minutes, so I might be wrong!

Above said, I noticed that it uses generics and introduces `T extends ApnsPushNotification`. From the perspective of polymorphism this means that an instance of the class can't be used with different implementations of `ApnsPushNotification`. Is this really a deliberate decision? Proposal here is to make the individual methods generic, not the whole class in itself. That is, replacing with

    public Future<PushNotificationResponse<T>> sendNotification(final T notification)

with

    public <T extends ApnsPushNotification> Future<PushNotificationResponse<T>> sendNotification(final T notification)

. What do you think of this? Doable?

jchambers:
As suggested in #342, this change makes the `ApnsClient` class non-generic, but instead makes its `sendNotification` method generic. This makes an `ApnsClient` much more flexible in terms of the types of notifications it can handle, which is definitely a good thing. This does come with some costs, though:

1. There's some API breakage. Good thing we have a big "we may change the API before 1.0" disclaimer at the bottom of the README ;)
2. The sending code itself is incrementally more complex.
3. We have an additional listener in the cycle for each notification we send, which causes some slight (~1%) performance decreases as shown below:

```
# Before

(flushImmediately)  (notificationCount)  Mode  Cnt  Score   Error  Units
              true                10000    ss   40  0.632 ± 0.022   s/op
             false                10000    ss   40  0.606 ± 0.018   s/op

# After

(flushImmediately)  (notificationCount)  Mode  Cnt  Score   Error  Units
              true                10000    ss   40  0.640 ± 0.017   s/op
             false                10000    ss   40  0.615 ± 0.020   s/op
```

In all, I think the tradeoff is worth it, but would very much appreciate other perspectives.

----

TODO:

- [x] Update inline docs
- [x] Update README

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -99,11 +99,9 @@ import io.netty.util.concurrent.SucceededFuture;
  *
  * @author <a href="https://github.com/jchambers">Jon Chambers</a>
  *
- * @param <T> the type of notification handled by the client
- *
  * @since 0.5
  */
-public class ApnsClient<T extends ApnsPushNotification> {
+public class ApnsClient {
     private final Bootstrap bootstrap;
     private volatile ProxyHandlerFactory proxyHandlerFactory;
     private final boolean shouldShutDownEventLoopGroup;
@@ -118,7 +116,7 @@ public class ApnsClient<T extends ApnsPushNotification> {
     private volatile ChannelPromise reconnectionPromise;
     private long reconnectDelaySeconds = INITIAL_RECONNECT_DELAY_SECONDS;
 
-    private final Map<T, Promise<PushNotificationResponse<T>>> responsePromises = new IdentityHashMap<>();
+    private final Map<ApnsPushNotification, Promise<PushNotificationResponse<ApnsPushNotification>>> responsePromises = new IdentityHashMap<>();
 
     private ApnsClientMetricsListener metricsListener = new NoopMetricsListener();
     private final AtomicLong nextNotificationId = new AtomicLong(0);
@@ -221,7 +219,7 @@ public class ApnsClient<T extends ApnsPushNotification> {
                     @Override
                     protected void configurePipeline(final ChannelHandlerContext context, final String protocol) {
                         if (ApplicationProtocolNames.HTTP_2.equals(protocol)) {
-                            final ApnsClientHandler<T> apnsClientHandler = new ApnsClientHandler.ApnsClientHandlerBuilder<T>()
+                            final ApnsClientHandler apnsClientHandler = new ApnsClientHandler.ApnsClientHandlerBuilder()
                                     .server(false)
                                     .apnsClient(ApnsClient.this)
                                     .authority(((InetSocketAddress) context.channel().remoteAddress()).getHostName())
@@ -385,7 +383,6 @@ public class ApnsClient<T extends ApnsPushNotification> {
             this.gracefulShutdownTimeoutMillis = timeoutMillis;
 
             if (this.connectionReadyPromise != null) {
-                @SuppressWarnings("rawtypes")
                 final ApnsClientHandler handler = this.connectionReadyPromise.channel().pipeline().get(ApnsClientHandler.class);
 
                 if (handler != null) {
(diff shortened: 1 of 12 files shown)


pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
```diff
@@ -393,7 +393,7 @@
      *
      * @since 0.8
      */
-    public ApnsClient<T> build() throws SSLException {
+    public ApnsClient build() throws SSLException {
```
panchenko: probably the whole ApnsClientBuilder class should not be generic.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -103,7 +103,7 @@
  *
  * @since 0.5
  */
-public class ApnsClient<T extends ApnsPushNotification> {
+public class ApnsClient {
```
panchenko: `<T>` should be removed from the class javadoc too.

panchenko:
@jchambers I think it is not necessary to have that additional listener, `SimplePushNotificationResponse` could be created directly, as after type erasure it is aware only of `ApnsPushNotification`.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
jchambers: Good catch! Done in a4c4957.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: Agreed—I'll be making a comprehensive pass at the docs shortly.

jchambers:
> I think it is not necessary to have that additional listener

I kept thinking the same thing when I was working on this, but kept struggling to make it work out in the end. Can you be more specific with your suggestion? I'm totally willing to believe (and strongly suspect) that I've missed something obvious.

panchenko:
> Can you be more specific with your suggestion?

I am thinking of having declaration like this:

`private final Map<ApnsPushNotification, Promise<PushNotificationResponse<ApnsPushNotification>>> responsePromises`

and the only cast needed is in `sendNotification()` when adding the new promise there.

jchambers:
@panchenko That was pretty much my thought, too. Ultimately, we have to do something like this:

```java
ApnsClient.this.responsePromises.put(notification, (Promise<PushNotificationResponse<ApnsPushNotification>>) responsePromise);
```

…but then the compiler complains:

> Cannot cast from `Promise<PushNotificationResponse<T>>` to `Promise<PushNotificationResponse<ApnsPushNotification>>`

…and that's the part I haven't quite figured out yet. I'm not sure why we can't make that cast; it seems like we know that, by definition, `T extends ApnsPushNotification`. I'll admit that I haven't yet sat down with a pen and paper to work out the erasure implications, but this seems to be the fundamental limitation here.

jchambers:
Good call. When in doubt, be vague ;)

Done in e0698cd, and I think that helps things quite a bit.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -493,8 +492,8 @@ public void run() {
 
                                 @Override
                                 public void run() {
-                                    for (final Promise<PushNotificationResponse<T>> responsePromise : ApnsClient.this.responsePromises.values()) {
-                                        responsePromise.tryFailure(new ClientNotConnectedException("Client disconnected unexpectedly."));
+                                    for (final Promise<PushNotificationResponse<ApnsPushNotification>> partialResponsePromise : ApnsClient.this.responsePromises.values()) {
+                                        partialResponsePromise.tryFailure(new ClientNotConnectedException("Client disconnected unexpectedly."));
```
panchenko: I would suggest renaming `partialResponsePromise` back to `responsePromise` here.

jchambers: Good catch—got it in 75c9fe6.

PR #350, opened 2016-07-28
Roll back fancy flush thresholds (closes #332)

### LINKED ISSUE #332: [0.7.3] setFlushThresholds(0, 0) make sending notification delayed
Test code:

```java
public class PushyTest {
    public static void main(String[] args) throws Exception {
        ApnsClient<SimpleApnsPushNotification> apnsClient;
        apnsClient = new ApnsClient<>(...);
        apnsClient.setFlushThresholds(0, 0);

        apnsClient.connect(ApnsClient.DEVELOPMENT_APNS_HOST).await();
        System.out.println("connected");

        // Wait event loop to start polling
        Thread.sleep(1000);

        SimpleApnsPushNotification notification = new SimpleApnsPushNotification(...);
        System.out.println("send begin");
        apnsClient.sendNotification(notification).await();
        System.out.println("send complete");
        apnsClient.disconnect().await();
    }
}
```

It takes very long time between `send begin` and `send complete`. If changed to `setFlushThresholds(1, 1)`, it is sent immediately.

Stack dump right after `send begin` shows NioEventLoop blocked at `select()` call.

jchambers:
Flush thresholds are, I think, a good idea from a performance perspective, but have been causing some problems in practice (see #332). At the same time, it looks like a better, more robust, upstream solution is on its way (see #326). This pull request rolls back flush thresholds entirely with the understanding that we'll have a drop-in replacement soon.

So why do this? Why not just fix the underlying bug in #332? We could, but auto-flushing (#326) will represent a breaking API change when it happens, and v0.8 is loaded with API-breaking changes anyhow. The idea is to get the API breakage out of the way now, then chase it with a non-breaking performance boost later.

I am very, very open to arguments that this is a dumb idea and we should just fix the bug now ;)

### DIFF
--- a/benchmark/src/main/java/com/relayrides/pushy/apns/ApnsClientBenchmark.java
+++ b/benchmark/src/main/java/com/relayrides/pushy/apns/ApnsClientBenchmark.java
@@ -4,7 +4,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang3.RandomStringUtils;
 import org.openjdk.jmh.annotations.Benchmark;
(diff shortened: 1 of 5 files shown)


jchambers:
Discussion upstream makes it sound like it'll be a while before auto-flushing becomes A Thing. Considering our options in the meantime.

jchambers:
After more thought, I still think this is the right way to go. In short, I think this feature introduces a lot of complexity by trying to do something at the application level that should probably be happening inside Netty. For now, I think it makes sense to favor reliability and simplicity over performance.

PR #362, opened 2016-08-20
Move docs from `gh-pages` to `master`

jchambers:
Now that GitHub allows [docs to be serverd from `master` instead of a separate `gh-pages` branch](https://github.com/blog/2228-simpler-github-pages-publishing), I figure we should take advantage.


jchambers:
On further reflection, I actually don't think this is such a hot idea. Closing for now.

PR #364, opened 2016-08-23
Ability to disable the auto-reconnect feature

dimitark:
Added the ability to disable the auto-reconnect feature and the ability to add a disconnect listener to the client.

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -119,6 +119,16 @@ public class ApnsClient {
     private final AtomicLong nextNotificationId = new AtomicLong(0);
 
     /**
+     * If true, the client will try to reconnect automatically after an unexpected failure.
+     */
+    private boolean autoReconnect = true;
+
+    /**
+     * A disconnect listener.
+     */
+    private ApnsClientDisconnectListener disconnectListener;
+
+    /**
      * The default write timeout, in milliseconds.
      *
      * @since 0.6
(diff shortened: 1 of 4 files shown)


jchambers:
@dimitark Can you tell us a little more about the use case for this feature? When would you want to control client reconnection manually?

dimitark:
You don't always want to keep the connection alive indefinitely. At least not for all clients.

If you have a big number of clients and half of them are sending notifications frequently, but half of them are not (a few notifications every couple of days). You want to have the connections open all the time for the first half but you don't want to reconnect the second half after the service terminates the connection due to a timeout.

jchambers:
That makes sense, but I think we should pass on adding this feature all the same. Apple has announced an upstream change where we'll be moving away from connection-level authentication and toward per-notification authentication (see #314). When that's done, we'll be able to reuse one client for all topics, and that should eliminate the need for this feature.

If token authentication weren't on the horizon, I'd definitely think this was a good idea. Seem legit?

dimitark:
@jchambers I wasn't aware of the Token authentication feature. Nevertheless I think this feature is needed because we will definitely need to maintain both authentication methods in the foreseeable future, because I'm sure not all clients will jump to Token authentication right away and we will be stuck with the connection-level authentication for some time, which will cause unwanted open connections.

jchambers:
@dimitark With sincere apologies for the outrageously long delay, I think this is a ball better left in the caller's court. Callers ought to know how much traffic they're sending to various topics (especially when armed with metrics listeners), and it seems like resource management needs are likely to be sufficiently diverse that we're unlikely to be able to do a thing that makes everybody happy within Pushy.

With great respect and sincere thanks for the contribution, I think we're going to pass on this one for now.

PR #368, opened 2016-09-08
Add support for `subtitle` and `mutable-content` payload fields

jchambers:
This builds on the work by @ScienJus in #322. Beyond the initial changes (which added `subtitle` and `mutable-content`), this adds some docs and adds support for localized subtitles. Unfortunately, none of this is officially documented yet, so we just have to take our best guess.


jchambers:
@ScienJus look good?

ScienJus:
@jchambers  look good to me

PR #383, opened 2016-11-15
Add a backpressure mechanism for channel writes

stepheno:
* Checks the status of `Channel.isWritable()` before attempting to write a
notification to the channel. If `isWritable()` returns false, return a new failed
future with a cause of `ClientBusyException`. The presence of this exception is
a signal to back off write attempts until the underlying channel has sucessfully returned
to its low water mark
* Adds `ApnsClientBuilder#setChannelWriteBufferWatermark` for setting
 the channel watermark levels
* Adds `ApnsClientBuilder#setByteBufferAllocator` for setting the buffer
 allocator of the netty channel

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
@@ -34,6 +34,8 @@ import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLException;
 
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.WriteBufferWaterMark;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -85,6 +87,9 @@ public class ApnsClientBuilder {
 
     private Long gracefulShutdownTimeout;
     private TimeUnit gracefulShutdownTimeoutUnit;
+
+    private ByteBufAllocator byteBufferAllocator;
+    private WriteBufferWaterMark channelWriteBufferWaterMark;
 
     private static final Logger log = LoggerFactory.getLogger(ApnsClientBuilder.class);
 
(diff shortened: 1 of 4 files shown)


jchambers:
On the whole, this looks good to me! I'll give it a more detailed read as soon as I can (which will most likely be Friday). Thanks very much for the contribution!

stepheno:
Thanks for taking a look. I just noticed that you have a 0.8.2 release milestone planned. Mind if update to target that release instead(just changing the @since annotations in the comments)?

jchambers:
@stepheno I went back and forth on this a few times, but I think it makes sense to include in 0.8.2.

pushy/src/main/java/com/relayrides/pushy/apns/ClientBusyException.java
```diff
@@ -0,0 +1,49 @@
...
+    public ClientBusyException() {
+        super();
+    }
+
+    /**
+     * Constructs a new exception with the given message.
+     *
+     * @param message a short, human-readable explanation of the cause of this exception
+     */
+    public ClientBusyException(final String message) {
```
jchambers: It looks like we never use this constructor; could we just drop it? If all we're left with is the default passthrough constructor, we could probably drop that, too.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
```diff
@@ -349,6 +354,34 @@ public ApnsClientBuilder setGracefulShutdownTimeout(final long gracefulShutdownT
...
+     *
+     * @since 0.9
+     */
+    public ApnsClientBuilder setByteBufferAllocator(ByteBufAllocator byteBufferAllocator) {
+        this.byteBufferAllocator = byteBufferAllocator;
+        return this;
+    }
+
+    /**
+     * Sets the buffer usage watermark range for the underlying netty channel
```
jchambers: Can we be a little more descriptive here, too? This seems like a good place to mention the existence of `ClientBusyException`, for example.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientBuilder.java
```diff
@@ -349,6 +354,34 @@ public ApnsClientBuilder setGracefulShutdownTimeout(final long gracefulShutdownT
     }
 
     /**
+     * Sets the buffer allocator for the underlying netty channel
```
jchambers: Can we be a little more descriptive here and maybe explain why a user would want to do this, or what the consequences might be? Also, as a super minor thing, mind adding a period to the end of the sentence? ;)

jchambers:
Actually, @stepheno, the things I'm asking for here are pretty minor. I'll go ahead and make the doc changes myself. Please consider yourself off the hook. I'll cobble a thing together and merge/close this shortly. Thanks again!

stepheno:
Hey Jon,

Sorry for the slow response, been a bit busy at work this week. Thanks for
updating the PR.

Also, looking forward to the upcoming token auth changes.

Cheers

jchambers:
Hey, @stepheno, can you tell me more about the use case for setting a custom byte buffer allocator? To my knowledge, the only options are "pooled" and "unpooled," and those can be selected globally by setting the `io.netty.allocator.type` system property:

```sh
java -Dio.netty.allocator.type=pooled -jar totally-sweet-push-notification-thing.jar
```

Do you have a use case in mind where somebody (a) would need a non-default allocator and (b) would need a different one than the JVM-wide default? To be clear, I'm totally open to this, but I want to double-check that the need isn't already met by the `-Dio.netty.allocator.type` thing.

stepheno:
Hey @jchambers, you're right about the global `io.netty.allocator.type` system property. There's not really a need to add functionality to the builder for this case. Feel free to rip it out.

jchambers:
And we're good. I did the cleanup stuff in 5882679; please let me know if anything looks amiss to you. Cheers!

jchambers:
Er… by which I actually mean 1e221a3e799406fd43d06cc2e8695b75d48967cc.

Issue #399, opened 2016-12-13
Add connection pooling


jchambers:
With TLS-based client authentication, clients could only send notifications to the topics mentioned in the certificate used for authentication. With the arrival of token-based authentication, a single client can send notifications to any number of topics.

One consequence is that many users may now want to centralize on a single client instance for simplicity. For performance reasons, we should consider allowing clients to operate a pool of connections to make better use of CPU resources (right now, a client will use exactly one thread regardless of the number of cores available) and possibly upstream resources (Apple claims that it's beneficial to spread connections across multiple servers, but I'm skeptical).

jchambers:
I took a first shot at connection pooling in the [`connection_pooling_spike`](https://github.com/relayrides/pushy/compare/drop_tls_authentication...connection_pooling_spike) branch and have some observations to share around (a) design implications and (b) performance.

In terms of design changes, it seems like one thing we may want to give up on is the idea of explicitly "starting" an `ApnsClient`, and instead letting it lazily-create channels. In the past, this would have been a really bad idea because there were a lot of ways for connections attempts to fail without much hope of self-resolution:

- Callers might not have Apple's intermediate certificates in their trust store
  - This is less of an issue now that Apple is using a more widely-recognized CA, and more to the point, one that's in most JRE trust stores by default
- Callers might have an expired client certificate
  - Auth tokens don't expire, but can be revoked; either way, this would be a problem when actually sending notifications, but not when establishing a connection
- Callers might have a screwy/malformed keystore
  - Now that callers don't have to deal with CSRs, exporting certificates, and [weird Keychain Access bugs](https://github.com/relayrides/pushy/wiki/Certificates), external factors are much less likely to interfere with the connection process

With token-based authentication, the things that can go wrong with opening a connection are mostly network-level things that are much more likely to self-resolve (or be beyond the control of the caller), and so it's less important that we provide a clear view into the status of the initial connection attempt.

On performance, some preliminary benchmarking shows per-channel throughput to be slower than the single-channel approach by about a factor of two. There are three things that I think are driving the slowdown:

1. Client channels need to deal with thread contention when working with the client's `responseFutures` map.
2. Client channels (and mock server channels) need to deal with thread contention when getting `AuthenticationTokenSuppliers`.
3. Actually acquiring and releasing channels now involves some kind of synchronization and is generally a new cost that we simply didn't have to pay before.

[...]

PR #403, opened 2016-12-18
Move response promise handling into the client handler

jchambers:
This moves response promise handling from the `ApnsClient` into the `ApnsClientHandler`. In the single-channel case, this doesn't make much of a difference, but it does remove an obstacle to multi-channel operation, as discussed in #399.

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -37,7 +37,6 @@ import java.security.spec.PKCS8EncodedKeySpec;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -51,6 +50,7 @@ import com.relayrides.pushy.apns.proxy.ProxyHandlerFactory;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
@@ -135,8 +135,6 @@ public class ApnsClient {
     private volatile ChannelPromise reconnectionPromise;
     private long reconnectDelaySeconds = INITIAL_RECONNECT_DELAY_SECONDS;
 
-    private final Map<ApnsPushNotification, Promise<PushNotificationResponse<ApnsPushNotification>>> responsePromises = new IdentityHashMap<>();
-
     private ApnsClientMetricsListener metricsListener = new NoopMetricsListener();
     private final AtomicLong nextNotificationId = new AtomicLong(0);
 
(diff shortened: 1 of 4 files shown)


pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -826,46 +802,34 @@ protected AuthenticationTokenSupplier getAuthenticationTokenSupplierForTopic(fin
...
-                        ApnsClient.this.responsePromises.remove(notification);
-                        responsePromise.tryFailure(future.cause());
+                            @Override
+                            public void operationComplete(final ChannelFuture future) throws Exception {
+                                if (future.isSuccess()) {
+                                    ApnsClient.this.metricsListener.handleNotificationSent(ApnsClient.this, notificationId);
+                                }
+                            }
+                        });
+                    } catch (final IllegalStateException e) {
```
jchambers: I'm debating making this a more purpose-specific exception. On one hand, it seems bogus to have a ton of super narrowly-scoped exceptions (maybe they should have common base class?), but on the other, we might get `IllegalStateExceptions` for other reasons.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -826,46 +802,34 @@ protected AuthenticationTokenSupplier getAuthenticationTokenSupplierForTopic(fin
...
-            connectionReadyPromise.channel().writeAndFlush(notification).addListener(new GenericFutureListener<ChannelFuture>() {
-
-                @Override
-                public void operationComplete(final ChannelFuture future) throws Exception {
-                    if (future.isSuccess()) {
-                        ApnsClient.this.metricsListener.handleNotificationSent(ApnsClient.this, notificationId);
-                    } else {
-                        log.debug("Failed to write push notification: {}", notification, future.cause());
+                    try {
+                        handler.setResponsePromiseForNotification(notification, responsePromise);
```
jchambers: I wish I could think of a way to not do this, or at least make it as part of the `write` operation. I'll think on it, but I'm very open to suggestions in the meantime.

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: @jmason 👍  to doing it or not doing it? ;)

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
jchambers: The best idea I've come up with to make this a little more atomic is to actually write some kind of `PushNotificationAndResponsePromise` object instead of just the push notification. I think the awkwardness there may outweigh the benefit, though.

jchambers: On second thought, I think we may *have* to combine the promise and the notification. We need some thread-safe way to register the promise, and what we're doing now isn't thread-safe. We can either synchronize, which will get expensive quickly, or we can just do everything in the event loop. Much as I dislike the "write a promise" idea from a conceptual perspective, I think practical concerns win out here.

jchambers: I am on crazy pills and we are already doing this in the event loop.

jchambers: Eh. Did it anyway!

PR #404, opened 2016-12-18
Move authentication token handling into the client handler

jchambers:
As discussed in #399, this moves token-handling machinery out of the central `ApnsClient` and into the `ApnsClientHandler`. For now, while we only have one handler per client, this just adds complexity. When we try to add multiple handlers per client, though, this will be a big help, since clients won't need to synchronize to get access to auth tokens (which they need to do *a lot*).

One concern with this approach: if we have multiple handlers, each handler will keep its own copy of each `AuthenticationTokenProvider`, which means that the memory required to store each signing key increases linearly with the number of handlers in play. This isn't a big deal for small-scale users, and prooooobably an acceptable trade-off between throughput and memory usage for industrial-scale users, but I definitely think it's worthy of discussion.

----

TODO:

- [ ] Restore existing tests
- [ ] Add tests for post-connection token changes
- [ ] Move token handling to handlers for the mock server, too


pushy/src/main/java/com/relayrides/pushy/apns/AuthenticationTokenSupplier.java
```diff
@@ -27,18 +27,23 @@
  */
 class AuthenticationTokenSupplier {
 
-    private final Signature signature;
```
jchambers: I'd really love to find a way to not hang onto this, but `Signature` isn't cloneable, and we definitely want to create a new `AuthenticationTokenSupplier` for each handler.

kdubb:
@jchambers Would it be not be more adaptable to replace your use of `ApnsRegistry` in the `ApnsClientHandler` a simple interface for locating an `ApnsKey` for a specific topic; e.g. `ApnsKeyLocator`.  That coupled with some method of providing a specific implementation of `ApnsKeyLocator` to the `ApnsClientHandler` would allow users to match implementations to their usage.

I can think of 3 implementations people might need:
* **`ApnsKeyRegistry`**: A **_synchronized_** locator that can be shared between handlers; the registry name implies pre-registered keys. (i.e. the current method)
* **`ApnsConcurrentKeyRegistry`**: A locator that manages one `ApnsKeyRegistry` per handler (eschewing locking); against using. (i.e. the method proposed)
* **`ApnsOnDemandLocator`**: A locator that dynamically looks up keys as they are requested. An implementation based on the Guava `LoadingCache` is imagined.

To be implemented most flexibly you would need two interfaces:
* **`ApnsKeyLocator`**: The interface `ApnsClientHandler` uses to locate keys for specific topics.
* **`ApnsKeyLocatorFactory`**: The interface provided to `ApnsClient` used to create a `ApnsKeyLocator` for each new handler.

kdubb:
Proposed interfaces for the above:
```java
interface ApnsKeyLocator {
  /*
    Looks up an ApnsKey for a given topic.
    
    Providing the ApnsClientHandler during locating allows implementations to 
    optimize based on the individual handler (e.g. for concurrency).
   */
  ApnsKey locate(ApnsClientHandler handler, String topic);
}

interface ApnsKeyLocatorFactory {
  /*
    Creates, shares or recycles an ApnsKeyLocator for the handler.

    Not sure these parameters are necessary, or even useful, for actual implementations
    but erring on the side of flexibility and providing as much context as possible.
   */
  ApnsKeyLocator create(ApnsClient client, ApnsClientHandler handler);
}
```

kdubb:
One final thought:

For "compatibility" you could make `ApnsClient` which would work with nothing but `ApnsKeyLocatorFactory` and be the most flexible implementation.  A `SimpleApnsClient` would manage an internal `ApnsKeyRegistery` (imagined above); providing all the same key registration methods `ApnsClient` currently does.

jchambers:
Hey, @kdubb, thanks for the suggestions! I'll mull them over, but it's worth pointing out that "adaptability" isn't necessarily a goal here; a lot of the new machinery that's getting introduced here—`ApnsKey`, `ApnsKeyRegistry`, `AuthenticationToken`, etc.—are all package-private classes (unless I've made a typo, which is 100% possible), and aren't intended for public consumption. If all goes well, we'll significantly change the inner workings of `ApnsClient` with callers never even knowing that anything changed (except for the part where they can assign more threads to a client to get more throughput).

kdubb:
@jchambers Yes I see they are currently internal.  In my opinion they need to be opened up so end users can participate in the decision making post-facto.

In your comments above you specifically talk about "small-scale" and "large-scale" users; stating you are attempting to be useful to both sets.  As we have built what would be considered a "large-scale" use (a platform that hundreds to thousands of different applications share) I can say that currently our biggest problem is having to pre-register keys for topics and not being able to cycle them in & out of a cache. In other words, the `ApnsOnDemandLocator` imagined above is something we'd spend considerable time perfecting our own version of, for our specific application.

You speak of throughput above but honestly we would gladly trade the throughput of Pushy for a shared locator that can load & cache keys on demand. This would allows us to carry the currently active set of keys in memory rather that pre-load all possible keys (which we are essentially doing now [over time]).  These features would provide greater overall throughput & predictability to our application as a whole rather than just increasing the throughput of Pushy itself.

The interfaces I laid out were a suggestion to truly allow both large-scale and small-scale users to be able to use Pushy without alteration. It's an awesome project!

jchambers:
Ah—I see. Sounds like there's a larger user need here, but I don't think this pull request is the right place to take on what you're suggesting.

Would you mind opening a separate issue with some more details about your use case? I'm particularly surprised to hear that—even with thousands of keys—key registration is your biggest challenge, and I'd love to learn more about that.

Thanks!

kdubb:
@jchambers I shall open a separate issue with these ideas and added details. Please realize I was referring to one specific aspect when I said "biggest problem"; Pushy works very well in most other aspects.

jchambers:
Totally understood, and I appreciate the kind words. I mean only that your use case stretches my mental model a bit, which is probably a good thing!

jchambers:
Regrettably, I think we have to shelve this for now. I've learned that the intended behavior is, in fact, "one team per connection." We'll probably salvage a lot of the good parts of this pull request (like the clearer model for keys/tokens), but the need for concurrent access to a key registry (or a even having a key registry in the first place) is now largely gone.

Issue #407, opened 2017-01-03
Back-pressure mechanism causes unexpected write exceptions


qiankunli:
there is a lot of "write failure" when sending a large amount of notifications. 

for ApnsClient class, responsePromises save the notifications that have not receive response, can you add a logic in sendNotification method：block sendNotification when the size of the responsePromises bigger than the threshhold(setting by user)。

jchambers:
It sounds like you're asking us to roll back #383, which I'm beginning to think we may have implemented a bit hastily.

@stepheno I think one weakness with the current approach is that there's no way for callers to know when the problem (too much stuff in the write queue) has been resolved. I have two thoughts:

1. Just remove `ClientBusyException` entirely. Callers can increment a counter when they send a notification, then add a listener that decrements the counter when a notification is resolved. As near as I can tell, that would meet the same user need without creating quite so much of an imposition on callers who aren't worried about flooding.
2. Replace `ClientBusyException` with a listener interface callers can implement to learn about client writability changes.

Since the main goal for the next release is to implement connection pooling, my inclination is to take the former approach. @stepheno would that be disastrous for your use case?

stepheno:
@jchambers I'm fine with the first approach. I can take a ticket/lease rate limiting semaphore approach in the interim. Looking forward to support for tokens and connection pooling :+1:

jchambers:
> Looking forward to support for tokens and connection pooling

Sanity check: we [already released](https://github.com/relayrides/pushy/releases/tag/pushy-0.9) auth token support. Are you looking for something different?

stepheno:
That's great news. I wasn't following that closely over the holidays, entirely missed the new release. Thanks for the heads up.

jchambers:
This has been rolled back in 3f45633.

PR #413, opened 2017-01-15
Cancel scheduled reconnect task during disconnecting ApnsClient

ylgrgyq:
Hi, Dear jchambers, 
I think we should cancel the scheduled reconnect task during disconnecting ApnsClient, otherwise we could have a leaking APNs connection which may never be closed. WDYT?

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -76,6 +76,7 @@ import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.GlobalEventExecutor;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.concurrent.SucceededFuture;
+import io.netty.util.concurrent.ScheduledFuture;
 
 /**
  * <p>An APNs client sends push notifications to the APNs gateway. Clients authenticate themselves to APNs servers in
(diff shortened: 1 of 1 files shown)


jchambers:
Good catch! I think this is important, but is likely to become less relevant after we ship 0.10 (which will include connection pooling and largely discard the notion of "reconnecting"). I think we should *also* fix #407 before 0.10 goes out, too, so I'll create an 0.9.1 branch shortly. When that happens, could I trouble you to re-target this pull request there?

ylgrgyq:
I'd love to, please ping me when you need me to do that. Thanks.

jchambers:
The `v0.9.1` branch is now A Thing; would you mind retargeting/rebasing at your convenience? Thanks!

ylgrgyq:
Done. PTAL @jchambers

jchambers:
(FYI: you can just change the target branch by editing the pull request)

pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
```diff
@@ -469,7 +471,7 @@ public void operationComplete(final ChannelFuture future) throws Exception {
                                 if (ApnsClient.this.reconnectionPromise != null) {
                                     log.debug("Disconnected. Next automatic reconnection attempt in {} seconds.", ApnsClient.this.reconnectDelaySeconds);
 
-                                    future.channel().eventLoop().schedule(new Runnable() {
+                                    scheduledReconnectFuture = future.channel().eventLoop().schedule(new Runnable() {
```
jchambers: Minor to the point that I almost feel bad asking: could you please qualify this with `ApnsClient.this`?

ylgrgyq: Ah, yes, no problem

jchambers: Thanks! Will merge as soon as tests pass.

PR #414, opened 2017-01-18
Possible concurrent modify on a plain HashMap

ylgrgyq:
I notice in the comment on method [registerSigningKey](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java#L556) of class ApnsClient says that ["Tokens may be registered at any time in a client's life-cycle"](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java#L541). So when one thread is registering a signing key to a ApnsClient while there could be another thread is retrieving `AuthenticationTokenSupplier` from the same ApnsClient by calling [getAuthenticationTokenSupplierForTopic](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java#L754).  Because HashMap is not thread safe so concurrent modifying it and reading it could lead to some unexpected behavior. So may be we could change `authenticationTokenSuppliersByTopic` and `topicsByTeamId` to ConcurrentHashMap. WDYT? 

If this PR is OK, I can cherry-pick this commit to the branch v0.9.1 and give another PR targeting v0.9.1.

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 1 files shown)


jchambers:
Yep; that's certainly a mistake on my part. I don't think this will make sense in 0.10 (all of this stuff will move into `ApnsClientHandler`, so thread safety won't be an issue in quite the same way), but let's plan on getting it in to 0.9.1. I'd like to run some benchmarks before merging this to make sure it's not going to cause a huge performance decrease (I don't expect it will), but it looks good to me in principle. As always, thanks for the contribution!

jchambers:
Cherry-picked this into v0.9.1 as 8161952. Thanks very much!

PR #440, opened 2017-03-31
Update to Netty 4.1.9/netty-tcnative 2.0.0

jchambers:
This updates Netty and netty-tcnative to their most recent versions. Perhaps most significantly, the update to netty-tcnative resolves the longstanding conflict between netty-tcnative and Tomcat's own (original) tcnative implementation, which makes life *much* easier for Tomcat users.


jchambers:
One very significant change in here: we're now relying on `netty-tcnative-boringssl-static` by default. This should make things work better out of the box for nearly all users. We've provided instructions for switching to a JDK SSL provider for users who can't use `netty-tcnative-boringssl-static`, but they should be few and far between. Users who want a statically-linked native SSL provider other than boringssl probably know what they're doing already ;)

matzew:
sweet

Issue #433, opened 2017-03-17
 Closing connection due to write timeout


yi719:
Hi,

sometimes send notifications in production environment will got errors like this:

```
DEBUG c.r.pushy.apns.ApnsClientHandler    Closing connection due to write timeout.
DEBUG com.relayrides.pushy.apns.ApnsClient    Disconnected. Next automatic reconnection attempt in 1 seconds.
DEBUG i.n.h.c.http2.Http2ConnectionHandler    [id: 0x4699b554, L:/10.10.91.99:54455 ! R:api.push.apple.com/17.188.164.209:443] Sending GOAWAY failed: lastStreamId '0', errorCode '2', debugData ''. Forcing shutdown of the connection.
javax.net.ssl.SSLException: SSLEngine closed already
    at io.netty.handler.ssl.SslHandler.wrap(...)(Unknown Source) ~[netty-handler-4.1.5.Final.jar:4.1.5.Final]
DEBUG com.relayrides.pushy.apns.ApnsClient    Failed to write push notification: SimpleApnsPushNotification [token=, payload={"aps":{"content-available":1},"refer_url":"","seq":17002}, invalidationTime=null, priority=IMMEDIATE, topic=, apns-collapse-id=null]
java.nio.channels.ClosedChannelException: null
    at io.netty.channel.AbstractChannel$AbstractUnsafe.close(...)(Unknown Source) ~[netty-transport-4.1.5.Final.jar:4.1.5.Final]
DEBUG i.n.h.c.http2.Http2ConnectionHandler    [id: 0x4699b554, L:/10.10.91.99:54455 ! R:api.push.apple.com/17.188.164.209:443] Sending GOAWAY failed: lastStreamId '0', errorCode '2', debugData ''. Forcing shutdown of the connection.
javax.net.ssl.SSLException: SSLEngine closed already
    at io.netty.handler.ssl.SslHandler.wrap(...)(Unknown Source) ~[netty-handler-4.1.5.Final.jar:4.1.5.Final]
DEBUG com.relayrides.pushy.apns.ApnsClient    Failed to write push notification: SimpleApnsPushNotification [token=, payload={"aps":{"content-available":1},"refer_url":"","seq":16002}, invalidationTime=null, priority=IMMEDIATE, topic=, apns-collapse-id=null]
java.nio.channels.ClosedChannelException: null
[...]

PR #444, opened 2017-04-06
Drop write timeout

jchambers:
As discussed in #433, it seems increasingly clear that a write timeout—or at least how we're reacting to a write timeout—isn't very helpful. This change drops write timeouts entirely; we'll replace them with something more helpful in a follow-up pull request.

Two options for a replacement come to mind:

1. We could add a high/low water mark system for unresolved push notifications. We'd allow callers to register listeners when the water marks are crossed, but wouldn't take action beyond that (instead allowing callers to decide how and when to back off).
2. We could add per-notification timeouts. This would be a little more complex, and could lead to weird situations where we'd say a notification had timed out, but it ultimately got accepted by the server.

Opinions from the audience?

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClient.java
@@ -30,7 +30,6 @@ import io.netty.handler.ssl.ApplicationProtocolNames;
 import io.netty.handler.ssl.ApplicationProtocolNegotiationHandler;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.timeout.IdleStateHandler;
-import io.netty.handler.timeout.WriteTimeoutHandler;
 import io.netty.resolver.DefaultAddressResolverGroup;
 import io.netty.resolver.NoopAddressResolverGroup;
 import io.netty.util.concurrent.*;
@@ -96,7 +95,6 @@ public class ApnsClient {
 
     private final ApnsSigningKey signingKey;
 
-    private long writeTimeoutMillis = DEFAULT_WRITE_TIMEOUT_MILLIS;
     private Long gracefulShutdownTimeoutMillis;
 
     private volatile ChannelPromise connectionReadyPromise;
@@ -106,13 +104,6 @@ public class ApnsClient {
 
     private ApnsClientMetricsListener metricsListener = new NoopMetricsListener();
     private final AtomicLong nextNotificationId = new AtomicLong(0);
-
-    /**
-     * The default write timeout, in milliseconds.
-     *
-     * @since 0.6
-     */
-    public static final long DEFAULT_WRITE_TIMEOUT_MILLIS = 20_000;
 
     /**
      * The hostname for the production APNs gateway.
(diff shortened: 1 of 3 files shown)


jchambers:
Oops—almost forgot. The other thing we're doing is getting a little more aggressive with ping timeouts. This should help us detect truly dead connections without punishing users for having a backlog of unresolved notifications.

jmason:
What we do is have two levels of timeout; one is a per-request timeout, where a failure to deliver the push notification within the timeout increments an error counter and the push notification is considered not delivered (although it may actually have made it through); and the other is an overall batch timeout, where we stop attempting delivery of any of the remaining push notifications in the batch if it expires.   Effectively I think the former (the per-request timeout) is the per-notification timeout you are suggesting, and it seems to work ok in practice.  We're using 60 seconds for this.

The initial SSL connection is retried, and we retry on a ClientNotConnectedException, but anything from the sendNotification() onwards is not.

jchambers:
Thanks, @jmason! Out of curiosity, how often do you see individual notifications time out? Is it a constant, low-level background noise sort of thing, or do you generally see big spikes that represent a server hiccup?

jmason:
Low-level background noise is about right -- it looks like over the past week (for example) it's averaged around 10-20 per hour, which is infinitesimal compared to the volume of traffic.

jchambers:
Cool. Thank you!

jchambers:
I'm increasingly confident that the root cause of #433 is HTTP/2 stream saturation. If that's the case, and if we want to try per-notification timeouts, we'd need to make sure we're closing streams when we time out notifications. I need to root around in the Netty source code (particularly [`StreamBufferingEncoder`](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/StreamBufferingEncoder.java)) to make sure that will actually work if the network is super flaky. The concern (which I am totally making up and not basing on an actual reading of any spec or source code) is that we need to wait for the server to acknowledge that a stream has been closed, and if the underlying problem is that network conditions prevent us from receiving *anything* from the server, then we're not really gaining anything by closing streams locally.

ylgrgyq:
I think this PR can really fix #433. I merged this PR and deployed it on our server. Then I haven't seen exceptions mentioned in #433 any more. The TCP RST is gone too.

Thanks a lot for your work!

Now I'll keep my eyes on our server a little while longer to make sure there are no other problems.

codesnowstorm:
> As discussed in #433, it seems increasingly clear that a write timeout—or at least how we're reacting to a write timeout—isn't very helpful. This change drops write timeouts entirely; we'll replace them with something more helpful in a follow-up pull request.
> 
> Two options for a replacement come to mind:
> 
> 1. We could add a high/low water mark system for unresolved push notifications. We'd allow callers to register listeners when the water marks are crossed, but wouldn't take action beyond that (instead allowing callers to decide how and when to back off).
> 2. We could add per-notification timeouts. This would be a little more complex, and could lead to weird situations where we'd say a notification had timed out, but it ultimately got accepted by the server.
> 
> Opinions from the audience?
——————————
Hello！Has the high/low water mark system been built？When connection is unavailable, I need writeTimeOut feature here.Between performance and availability, I focus more on availability.

PR #445, opened 2017-04-06
README update: properly handle ApnsServerException

justinpermar:
Add to README in light of https://groups.google.com/forum/#!topic/pushy-apns/-Lwg9fhHKOs. I di not update Javadoc, which may be a good idea: http://relayrides.github.io/pushy/apidocs/0.9/com/relayrides/pushy/apns/ApnsClient.html#sendNotification-T-


jchambers:
Hi, @justinpermar! Please accept my apologies for the slow reply; we appreciate the contribution, but ultimately, the suggested change here isn't quite correct. In the event of a server error, waiting for reconnection won't necessarily help since the client doesn't automatically disconnect when there's an upstream error.

Respectfully, I think we must pass on this pull request, but we look forward to your future contributions. Cheers!

Issue #452, opened 2017-04-18
Configurable Idle Time Settings


kingster:
In our production environment, our corporate firewall has much lesser idle timeout settings(causing connection drops, which can be avoided by setting correct idle_time), and currently pushy has the following settings as hard coded values

```java
static final int PING_IDLE_TIME_MILLIS = 60_000; // milliseconds
```

Does it make sense to set this via the ApnsClientBuilder?

jchambers:
Sure. Why not?

kingster:
Just had one question, In `ApnsClientHandler` there is this

```java
private static final int PING_TIMEOUT = 30; // seconds
```

As far as I understand this is the PING reply timeout, and there is an assert 

```java
assert PING_TIMEOUT < ApnsClient.PING_IDLE_TIME_MILLIS;
```
Right now PING_TIMEOUT has been set as half of idle time, should this be the scenario always?

kingster:
Ah, that assert is also buggy! It's comparing seconds with milliseconds! :(

jchambers:
> Right now PING_TIMEOUT has been set as half of idle time, should this be the scenario always?

`PING_TIMEOUT` must be less than `PING_IDLE_TIME`; we want to make sure that the "hey, the ping timed out" task appears in the event queue before the "fire the next `PING` frame" task. There are many roads we can take to get there. A mathematical relationship between the two seems reasonable if we're allowing callers to specify `PING_IDLE_TIME`; letting callers specify both seems needlessly complicated and error-prone.

> Ah, that assert is also buggy! It's comparing seconds with milliseconds!

Yes. This is already fixed in another branch.

jchambers:
Oops—we merged #453 a few days ago, which closes this issue. Thanks again for the contribution!

Issue #448, opened 2017-04-13
metricsListener notificationAccepted handler is called on ExpiredProviderToken rejection


jpermar:
It appears that Pushy metricsListener notificationAccepted handler is called when a push to Apns response has rejectionReason=ExpiredProviderToken. Here is what I see:

1) send a notification through Pushy

2) ApnsClient debug messages report two notifications sent: one with response "success=false, rejectionReason=ExpiredProviderToken" and second with "success=true, rejectionReason=null".

3) the notification sent handler is called twice, as expected in this case due to the internal Pushy retry, but the notificationAccepted handler is also called twice, even though only one response was "success=true".

jchambers:
@jpermar I'm having a hard time reproducing this. I added tests in 186240d, but they all pass. I think this may mean that we fixed the problem incidentally in #451, which untangled some token-handling logic.

I'm going to close this for now (since I think the problem is solved), but please do leave a comment if you feel that something more is required.

jchambers:
Looks like I spoke too soon. The [tests do fail](https://travis-ci.org/relayrides/pushy/jobs/229587531), but only sometimes. I think that means we have a race condition somewhere. Will continue to investigate.

jchambers:
Looks like the problem was with the tests and not the underlying client code. I think everything is under control as of 49f45e1. Thanks for your patience!

PR #453, opened 2017-04-18
Make idlePingInterval Configurable

kingster:
As discussed in #452

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientHandler.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientHandler.java
@@ -54,9 +54,7 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
     private final String authority;
 
     private ScheduledFuture<?> pingTimeoutFuture;
-
-    static final int PING_IDLE_TIME_MILLIS = 60_000; // milliseconds
-    private static final int PING_TIMEOUT_MILLIS = PING_IDLE_TIME_MILLIS / 2;
+    private long pingTimeoutMillis;
 
     private static final String APNS_PATH_PREFIX = "/3/device/";
     private static final AsciiString APNS_EXPIRATION_HEADER = new AsciiString("apns-expiration");
@@ -77,6 +75,7 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
     public static class ApnsClientHandlerBuilder extends AbstractHttp2ConnectionHandlerBuilder<ApnsClientHandler, ApnsClientHandlerBuilder> {
 
         private String authority;
+        private long idlePingIntervalMillis;
 
         public ApnsClientHandlerBuilder authority(final String authority) {
             this.authority = authority;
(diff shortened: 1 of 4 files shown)


pushy/src/main/java/com/relayrides/pushy/apns/ApnsClientHandler.java
```diff
@@ -361,7 +361,7 @@ public void operationComplete(final ChannelPromise future) throws Exception {
     @Override
     public void userEventTriggered(final ChannelHandlerContext context, final Object event) throws Exception {
         if (event instanceof IdleStateEvent) {
-            assert PING_TIMEOUT < ApnsClient.PING_IDLE_TIME_MILLIS;
+            assert PING_TIMEOUT < this.apnsClient.idlePingIntervalMillis;
```
kingster: @jchambers What should be the default value of PING_TIMEOUT? Currently, this assert will always pass since its comparing seconds < milliSeconds

jchambers: For folks following this thread, we've discussed this in https://github.com/relayrides/pushy/issues/452#issuecomment-294855725.

kingster:
Setting 
```java
pingTimeout =  Math.min(30_000, this.apnsClient.idlePingIntervalMillis/2)
```

So by default, it would be 30 seconds as it used to be. Incase idle timeout is less than 30 seconds, the timeout interval would be calculated as half of it. Let me know how you feel about this.

jchambers:
@kingster Please pardon the delay in getting to this pull request! This looks good, but could you please rebase/retarget against `master`?

Thanks!

kingster:
Have rebased the changes against the master. 

Not sure why travis is failing only for `openjdk7` with memory exception `
/usr/lib/jvm/java-7-openjdk-amd64/jre/lib/amd64/libnet.so(Java_java_net_Inet4AddressImpl_getLocalHostName+0x190)[0x7f0ac26f68e0]
[0x7f0ad1014e98]` Looks like compatibilty issue with new Precise enviroment

jchambers:
> Not sure why travis is failing only for openjdk7 with memory exception

Yeah—I ran into that last night with a README change, so I'm not super worried about it from a functional perspective. We've dealt with a similar problem in the past. I'll see what I can work out.

jchambers:
@kingster We restored the workaround for the Travis issue in #463. Would you mind rebasing one more time?

Thanks, and sorry for the trouble!

kingster:
No issues :)

Rebased it, but not sure why the test ` ApnsClientTest.testSendNotificationWithExpiredAuthenticationToken` is randomly failing on java8

```
https://travis-ci.org/relayrides/pushy/builds/229646755 (only j8, both)
https://travis-ci.org/relayrides/pushy/builds/229643685 (only j8, PUSHY_SSL_PROVIDER=jdk)
```

This was passing before. Any suggestions would be helpful

jchambers:
> Rebased it, but not sure why the test ApnsClientTest.testSendNotificationWithExpiredAuthenticationToken is randomly failing on java8

That's 100% my fault. See #448. Should have that cleaned up shortly. Sorry for the mess :\

jchambers:
Oooookay. I think everything should be under control as of 49f45e1. If you'd be so kind as to rebase just one more time, I think we'll be able to bring this to a close ;)

jchambers:
Thanks for your patience with all of the weird build hiccups, and thanks for the contribution!

PR #460, opened 2017-05-03
Update to Netty 4.1.10/netty-tcnative 2.0.1

jchambers:
This updates to the latest versions of Netty and netty-tcnative.


jchambers:
Huh… looks like something may be amiss upstream. Filed https://github.com/netty/netty/issues/6700.

jchambers:
Well, netty/netty#6700 has been fixed, but its existence does mean that Netty 4.1.10 is a no-go for us; we'll need to wait for 4.1.11.

PR #464, opened 2017-05-10
Use reference-counted native SSL provider

jchambers:
This change moves us to a reference-counted flavor of the native SSL provider when possible. This should reduce direct memory usage with no a priori downside.

### DIFF
--- a/pushy/src/main/java/com/relayrides/pushy/apns/SslUtil.java
+++ b/pushy/src/main/java/com/relayrides/pushy/apns/SslUtil.java
@@ -27,7 +27,7 @@ class SslUtil {
         if (OpenSsl.isAvailable()) {
             if (OpenSsl.isAlpnSupported()) {
                 log.info("Native SSL provider is available and supports ALPN; will use native provider.");
-                sslProvider = SslProvider.OPENSSL;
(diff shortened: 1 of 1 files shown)


jmason:
so does this avoid relying on the finalizer to perform off heap memory deallocation?  I ran into some problems with that recently, so it would be good to avoid that.

jchambers:
> so does this avoid relying on the finalizer to perform off heap memory deallocation?

Yes!

> I ran into some problems with that recently, so it would be good to avoid that.

To be clear, did you run into problems with using the finalizer or with *not* using the finalizer?

jmason:
Specifically, we ran into memory problems; we had a relatively large heap size, which was too close to the size of the host's RAM.  The effects of this were that a high throughput of off-heap OpenSSL context objects caused OS-level kernel OOM errors, because finalizers didn't get called in a timely manner (because heap-driven GC didn't happen frequently enough to trigger finalization frequently enough).

jchambers:
Cool. So, yeah—this should help ;)

Also, just spotted that [Netty 4.1.11 was just released](http://netty.io/news/2017/05/12/4-0-47-Final-4-1-11-Final.html), and that claims further fixes for `OPENSSL_REFCNT`. Will get that in shortly.

PR #481, opened 2017-06-15
Move channel closure logic from `close` to `channelInactive`

### LINKED ISSUE #479: ConcurrentModificationException when disconnecting client
Just found the following error in our logs

```
java.util.ConcurrentModificationException: null
        at java.util.IdentityHashMap$IdentityHashMapIterator.nextIndex(IdentityHashMap.java:734)
        at java.util.IdentityHashMap$ValueIterator.next(IdentityHashMap.java:832)
        at com.turo.pushy.apns.ApnsClientHandler.close(ApnsClientHandler.java:283)
        at io.netty.channel.AbstractChannelHandlerContext.invokeClose(AbstractChannelHandlerContext.java:624)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:608)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:465)
        at io.netty.channel.DefaultChannelPipeline.close(DefaultChannelPipeline.java:964)
        at io.netty.channel.AbstractChannel.close(AbstractChannel.java:234)
        at com.turo.pushy.apns.ApnsClient.disconnect(ApnsClient.java:618)
```

Maybe the access to the `responsePromises` map in `ApnsClientHandler` should be synchronized?

### LINKED ISSUE #480: Lost responses on reconnection
Seems that there is another problem with lost responses. It happens in our production environment about once per week. Monitoring system shows that couple thousand notifications do not receive response in short time and we manually cancel them (Future<PushNotificationResponse>) after some large timeout (it would be nice to have such a feature in pushy itself so caller simply gets some timeout or cancellation exception). I have limited access to production servers so can’t tell much details but it looks like some connection issues. Have tried to reproduce it locally, and seems that have found one case. If run tcpkill net 17.188.0.0/16 to simulate problems with network, in log I see how connections are dropped and then reconnection process starts. When then cancel tcpkill, everything restores and normal delivery process continues, but responses to some connections are missed - GenericFutureListener::operationComplete. is not called. From debug it seems that ApnsClientHandler::close is not called and therefore responsePromises are lost on channel close. May be responsePromises cleanup could be moved to some other place, seems that ApnsClientHandler::channelInactive would be fine.

Thanks,
Z.

jchambers:
This change moves the "fail all unresolved notifications" logic from the channel handler's `close` method to the `channelInactive` method. This should solve two problems at once:

1. `close` appears to be called only when we deliberately/cleanly close the channel, and it won't be called in cases of abnormal channel closure. `channelInactive`, on the other hand, should be called regardless. This fixes #480.
2. `close` can be called from outside of the channel's event loop, which means that past assumptions about serial executions were wrong and we could wind up with concurrent modification issues. Moving to `channelInactive` puts everything inside the event loop, which resolves #479.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,6 @@
-# Workaround for Travis CI issue 5227, which results in a buffer overflow
-# caused by Java when running the build on OpenJDK.
-# https://github.com/travis-ci/travis-ci/issues/5227
-before_install:
-    - cat /etc/hosts # optionally check the content *before*
-    - sudo hostname "$(hostname | cut -c1-63)"
-    - sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
-    - cat /etc/hosts # optionally check the content *after*
 language: java
+dist: trusty
+sudo: required
 jdk:
     - oraclejdk8
     - oraclejdk7
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
@@ -274,11 +274,11 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
     }
 
     @Override
-    public void close(final ChannelHandlerContext context, final ChannelPromise promise) throws Exception {
-        super.close(context, promise);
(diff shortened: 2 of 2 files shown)


jchambers:
The build failure here appears to be spurious, and I've updated https://github.com/travis-ci/travis-ci/issues/7732 accordingly. I'll try to find a workaround, but may just merge anyway if that doesn't work out.

Issue #476, opened 2017-06-07
Not all notifications got response


janzar:
Hi, 

Do anybody experience the problem, that each hour pushy does not get response to some sent notifications from Apple. That happens at the time when  auth token expires (it happens for about one hour and ten minutes when it is issued). Count of unanswered notifications depends on how many notifications are sent through connection. Seems that 50 notifications per second are enough.

It could be reproduced with following conditions. Say for example you have 8 threads where each tries to send 10 notifications at the same time. And for example you do it once a minute. When auth token expires, and you send next time, pushy receives response about expired token for some first notifications, but for all other remaining there is no response at all. It looks like that Apple when it receives new auth token in some way fails to respond on notifications with old token that it receives shortly after that. These numbers of threads and notifications I found empirically and with them this situation could be reproduced consistently. Also have tried to comment out auth token change and continue to send with expired token. on first attempt have not got response on all sent notifications, but on next iterations all notifications are responded. So it looks like a problem at Apple side, but would be nice if somebody can clarify. My pushy version is 0.10 and netty is 4.1.11.Final.

Thanks,
Z.

janzar:
P.S> when tested, I have sent all notifications to the same device.

janzar:
Hi @jchambers.

Could it be possible to add optional configurable time to live to TokenAuthenticationApnsClientHandler::authenticationToken. So when token is created, handler would schedule token recreation after some time. It could live together with current token expiration handling. Hope that then lost response problem would go away if token TTL is set less than one hour.

Seems that there is another problem with lost responses. It happens in our production environment about once per week. Monitoring system shows that couple thousand notifications do not receive response in short time and we manually cancel them (Future<PushNotificationResponse<ApnsPushNotification>>) after some large timeout (it would be nice to have such a feature in pushy itself so caller simply gets some timeout or cancellation exception). I have limited access to production servers so can’t tell much details but it looks like some connection issues. Have tried to reproduce it locally, and seems that have found one case. If run tcpkill net 17.188.0.0/16 to simulate problems with network, in log I see how connections are dropped and then reconnection process starts. When then cancel tcpkill, everything restores and normal delivery process continues, but responses to some connections are missed - GenericFutureListener::operationComplete. is not called. From debug it seems that ApnsClientHandler::close is not called and therefore responsePromises are lost on channel close. May be responsePromises cleanup could be moved to some other place, seems that ApnsClientHandler::channelInactive would be fine.

Thanks,
Z.

[...]

PR #484, opened 2017-06-24
Use Netty's stream management tools instead of rolling our own

jchambers:
In the process of working on #476, I noticed that Netty already has stream enumeration tools built into [`Http2Connection.Endpoint`](http://netty.io/4.1/api/io/netty/handler/codec/http2/Http2Connection.Endpoint.html). We should use those instead of rolling our own.

Changes best viewed with [whitespace-only changes hidden](https://github.com/relayrides/pushy/pull/484/files?w=1).


jchambers:
Build failures appear to be spurious.

xuwei-k:
> Build failures appear to be spurious.

https://github.com/travis-ci/travis-ci/issues/7964

PR #507, opened 2017-08-28
Correct dependency versions from pushy-parent pom

justinpermar:


README.md
```diff
@@ -23,10 +23,10 @@ If you use [Maven](http://maven.apache.org/), you can add Pushy to your project
 
 If you don't use Maven (or something else that understands Maven dependencies, like Gradle), you can [download Pushy as a `.jar` file](https://github.com/relayrides/pushy/releases/download/pushy-0.10/pushy-0.10.jar) and add it to your project directly. You'll also need to make sure you have Pushy's runtime dependencies on your classpath. They are:
 
-- [netty 4.1.12](http://netty.io/)
+- [netty 4.1.12.Final](http://netty.io/)
```
jchambers: Leaving the "final" out was actually a deliberate choice here. I absolutely agree that it's helpful for finding the right version in Maven Central, but (almost) by definition, the folks who are doing this manually aren't using Central. For human readability, I think it's best to just stick with the version number.

README.md
```diff
@@ -23,10 +23,10 @@ If you use [Maven](http://maven.apache.org/), you can add Pushy to your project
...
 - [netty-tcnative-2.0.3.Final](http://netty.io/wiki/forked-tomcat-native.html)
-- [gson 2.6](https://github.com/google/gson)
-- [slf4j 1.7.6](http://www.slf4j.org/) (and possibly an SLF4J binding, as described in the [logging](#logging) section below)
+- [gson 2.8.0](https://github.com/google/gson)
```
jchambers: Come to think of it, I'm not sure that we strictly _need_ GSON 2.8. We definitely needed 2.5 because it introduced some feature that wasn't in 2.4 (I can't for the life of me remember what, but will experiment to find it). I think requiring a recent version is the right thing to do in the pom (for bug fixes and performance improvements), but if users are putting things together by hand, it might make sense to point them to GSON 2.5 or newer. Let me think about that a little bit.

README.md
```diff
@@ -23,10 +23,10 @@ If you use [Maven](http://maven.apache.org/), you can add Pushy to your project
...
+- [slf4j 1.7.21](http://www.slf4j.org/) (and possibly an SLF4J binding, as described in the [logging](#logging) section below)
```
jchambers: We might just consider dropping the patch version here; anything in the 1.7 series is fine.

jchambers:
…also, if you wouldn't mind rebasing onto `master` so we can pick up the latest fixes for Travis' shenanigans, that'd be awesome!

jchambers:
@justinpermar Thanks again! I think we have this covered in cdc580e. Cheers!

PR #511, opened 2017-09-11
Support MDM

ricardopt:
With my pull request pushy will be able to handle MDM push notifications.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,3 @@
-# Workaround for Travis CI issue 5227, which results in a buffer overflow
-# caused by Java when running the build on OpenJDK.
-# https://github.com/travis-ci/travis-ci/issues/5227
-before_install:
-    - cat /etc/hosts # optionally check the content *before*
-    - sudo hostname "$(hostname | cut -c1-63)"
-    - sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
-    - cat /etc/hosts # optionally check the content *after*
 language: java
 jdk:
     - oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,11 @@ before_install:
 language: java
 jdk:
     - oraclejdk8
-    - oraclejdk7
     - openjdk7
 env:
     - PUSHY_SSL_PROVIDER=jdk
     - PUSHY_SSL_PROVIDER=native
 matrix:
     exclude:
-        - jdk: oraclejdk7
-          env: PUSHY_SSL_PROVIDER=jdk
         - jdk: openjdk7
           env: PUSHY_SSL_PROVIDER=jdk
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -27,11 +27,11 @@
     <parent>
         <artifactId>pushy-parent</artifactId>
         <groupId>com.turo</groupId>
-        <version>0.10.2</version>
+        <version>0.11-SNAPSHOT</version>
     </parent>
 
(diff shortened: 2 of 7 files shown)


jchambers:
I may be missing it, but what's different here? I get that you rearranged some methods, but it looks like they're still doing exactly the same thing as `buildWithMaximumLength`. Can you elaborate on the changes that MDM requires?

ricardopt:
I restructured some methods. The way that the build payload was being built it has "aps" as a mandatory key value. This doesn't work with MDN push notifications. You can see that I'm building the payload only with the key "mdn" and it has to be the only one in the payload. If you put i.e. "aps" and "mdn" it will not work.

I hope that the answer was clear if you need further information just ask. Mdn is a kind of black hole inside apple documentation.

jchambers:
> You can see that I'm building the payload only with the key "mdn" and it has to be the only one in the payload.

I think I'm still missing that, actually. Is it possible some of your changes got clobbered in a tragic merge accident? Also, could you please add unit tests for the behavior you're trying to achieve?

ricardopt:
I've added some unit testing.

It is easier if you go to files changed and then use the "View" hole file. When being there you can see that I introduce a private method that is the buildPayload that does the aggregation that was previously being made inside the buildPayloadWithMaximumLength, now it is used on the "common" and the "mdn". 

The unit-tests assert that the payload doesn't contain a key named "aps" and that the "pushMagic" value is assigned to the "mdn" key. The terminology is equal to the one used by the Apple documentation to be easier to implement by any Dev Team.

jchambers:
MDM is new to me and I'm catching up. If I'm understanding the docs correctly, though, the only time MDM and APNs intersect is when somebody wants to send a "wake up" message to the receiving device. In that case, the APNs payload always looks like this:

```json
{
    "mdm": "<push 'magic' value here>"
}
```

Is that also your understanding?

If so, I think we can greatly simplify the approach here by just adding a static method that takes a single "push magic" argument (see bfca46042cbcf980a91996f27011a55a032963c7). Seem legit?

ricardopt:
Yes. You said it all.

I don't know how the full pushy client works, I tried to maintain the same approach in the buffering that was being used in the other payload builders.

jchambers:
> Creating a direct branch instead of working on the pull request, the concept of open-source and community was a bit left in the rain there...

Friend, I'm open to feedback, but I think I need some clarification about your concerns. To be clear:

1. I haven't merged the changes in bfca460.
2. It seemed like showing you what I meant with actual code would be easier than explaining it in paragraph form.
3. In the same way that you can't push directly to `master`, I can't contribute directly to this pull request except by leaving comments.

With all that in mind, can you explain what you would have liked to be different in this exchange? Is your suggestion that I should have just described an alternative approach in comments, then left it to you from there?

ricardopt:
I think that you could put your suggestion on the comment. If the changes you made in the [bfca460](https://github.com/relayrides/pushy/commit/bfca46042cbcf980a91996f27011a55a032963c7)  were not merged you were left with two hypothesis:

- Clutter branches that mean nothing and are pure garbage;
- Lose context in the comments.

If I'm missing any hypothesis please add, but I don't see any other option that could arise from creating a new branch to answer a comment in a pull request.
---
Going back to the problem at hands, what do you think should be the approach? Keeping the uniformization of payload building or creating a static method that is slightly different from the current methodology.

Best regards,

ricardopt:
When are you thinking about making a new release with the MDM? We want to put our version in production and we are using my fork as a snapshot, we have to change that.

Thanks

jchambers:
Okay; let's roll with the static method, then. I'll close this pull request and open a new one. I'm thinking this will be out in 0.12, which does not yet have a scheduled release date.

Issue #417, opened 2017-01-25
MockApnsServer changes


ebabani:
Hi,

For our integration tests we've made some changes on the MockApnsServer to add the ability to see what pushes have been received, and record the topics/payloads of those pushes.

Do you have your own plans regarding something similar in the future, or would you be open to a PR?

kdubb:
This would be great... we resorted to a Node server that mocked the APNS server that essentially does the exact same thing.  Reusing the `MockApnsServer` would be helpful.

jchambers:
Can you elaborate a bit on the changes that you made? My inclination would be to add something like this:

```java
public interface MockApnsServerNotificationListener {
    void handleAcceptedNotification(ApnsPushNotification notification);
    void handleRejectedNotification(ApnsPushNotification notification, ErrorReason errorReason);
}
```

Seem legit? In any case, I'm kiiiiiinda taking a wrecking ball to `MockApnsServer` over in #404, so it might be a good idea to hold off on a pull request until that's a little less… wreck-y.

ebabani:
That's exactly the same approach I took. Add a notification listener, and record accepted/rejected notifications. 

The ApnsPushNotification object contains the http2 headers and payload so that the listener implementations can check whatever they wants.

Thanks for the heads up on #404 , I'll hold off until that's in.

jchambers:
I don't want to leave this hanging, but to be clear, #404 is temporarily on hold until we can get some clarification from Apple. Please see https://github.com/relayrides/pushy/issues/399#issuecomment-282306488 for additional discussion.

jchambers:
@ebabani With apologies for the long delay, I think the state of the server in `master` is pretty stable for the moment. We'd be thrilled to receive a pull request at your convenience!

jchambers:
For folks interested in this feature: do you mostly care that notifications made it to the server, or do you need something that runs notifications through the full validation gauntlet (verifying topics and auth tokens and such)?

One way I can think of to handle this is to have something like a `PushNotificationHandler` interface. Pushy could supply a `ValidatingPushNotificationHandler` that does all the checks we can derive from the APNs docs, and then end users could write something like a `JustTellMeIfAPushNotificationArrivedHandler` as needed. Things get a little more complicated if you folks need something that does all of the validation, but is easy to subclass. It's absolutely doable, but I'm just trying to strike the right balance of complexity and usefulness.

Thanks!

jchambers:
[...]

PR #526, opened 2017-10-02
Modular mock server

### LINKED ISSUE #459: Publicize benchmark server
As discussed in #450, it might be helpful for some users if we publicized the benchmark mock server. To do that well, I think we'll want the benchmark mock server and the existing mock server to share some abstract base class. That will be non-trivial, but probably worthwhile.

jchambers:
This work-in-progress change moves most of the business logic of the mock server into `PushNotificationHandlers` created by `PushNotificationHandlerFactories`. Pushy includes a few for common cases:

- `AcceptAllPushNotificationHandlerFactory` is useful for many unit tests where we're not interested in rigorous validation and for benchmarking.
- `ValidatingPushNotificationHandlerFactory` emulates the behavior of a real APNs server, and is useful for making sure clients are building push notifications correctly.

This indirectly closes #459 by way of doing away with the specialized benchmark server. The `AcceptAllPushNotificationHandlerFactory` is noticeably slower than the specialized server, but I think it's an acceptable trade-off given the gains in simplicity, reusability, and flexibility. In time, I think we can optimize things down to the point where the difference is minimal.

Another tradeoff: a handful of behind-the-scenes classes crept out into public view (I'm looking at you, `DateAsTimeSinceEpochTypeAdapter`). There aren't too many, though, and (again) I think it's a reasonable tradeoff.

I emphasize that this is a work in progress, but wanted to open the pull request early to solicit feedback. What do y'all think?

TODO:

- [x] Write unit tests for the validating handlers
- [x] Add a listener system as discussed in #417 (see #537)
- [x] Add a README or wiki page or something about using the mock server?


pushy/src/main/java/com/turo/pushy/apns/server/RejectionReason.java
```diff
@@ -0,0 +1,85 @@
...
+package com.turo.pushy.apns.server;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+/**
+ * An enumeration of reasons a push notification may be rejected by an APNs server. The most up-to-date descriptions of
+ * each rejection reason are available in Table 8-6 of Apple's <a href="https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html#//apple_ref/doc/uid/TP40008194-CH11-SW1">Local
+ * and Remote Notification Programming Guide - Communicating with APNs</a>.
+ */
+public enum RejectionReason {
```
jchambers: I'm really torn about how much to document in this enumeration. We've learned things that aren't really clear about some of the error codes from the (sparse) official documentation, but I'm hesitant to make strong statements of fact about things that we're still somewhat guessing about and that may change in the future.

I also feel weird about having this enum out in public view, but not using it in `PushNotificationResponse#getRejectionReason()`. The danger in using it there is that if Apple adds a new rejection reason, old versions of Pushy will break when they receive those new codes and can't map them to an enum key. We could have a catch-all `UNKNOWN` key, but that feels pretty gross. We could also add a `getRejectionReasonForErrorCode(String)` type of method, but that seems like so many extra steps that it wouldn't be worth it for most users.

Very open to ideas.

pushy/src/main/java/com/turo/pushy/apns/server/UnregisteredDeviceTokenException.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.UUID;
+
+/**
+ * An exception thrown by {@link PushNotificationHandler} instances to indicate that a push notification should be
+ * rejected by the server because its destination device token is no longer valid (presumably because the receiving app
+ * has been removed from the destination device).
+ *
+ * @since 0.12
+ */
+public class UnregisteredDeviceTokenException extends RejectedNotificationException {
```
jchambers: Felt weird to have `getDeviceTokenExpirationTimestamp` in the base `RejectedNotificationException` if only one error code could reasonably use it. Feels weird to have a special subclass for a single error code. I guess the alternative would be to have an exception for each kind of rejection, but that feels like overkill.

PR #535, opened 2017-10-06
Introduce an application-specific `PushNotificationFuture`

### LINKED ISSUE #439: Pushy should provide a way of getting a notification that couldn't be queued
The documentation suggests the best practice is:

```java

pushNotificationResponseHandler = 
   new FutureListener<PushNotificationResponse<SimpleApnsPushNotification>>() {
      public void operationComplete() {
           try {
                responseFuture.get()
           } catch (ClientNotConnectedException e) {
                  apnsClient.getReconnectionFuture.await()
                  //Resend!?
           }
      }
   };

future = apnsClient.sendNotification(pushNotification);
future.addListener(pushNotificationResponseHandler);
```

But there's no way to get a reference in a `FutureListener<PushNotificationResponse<SimpleApnsPushNotification>>` to the push notification which has failed to send.

Handling this case involves one of:

1. Making assumptions about Pushy (i.e. is the queue processed in order - can I just maintain my own queue of unsent notifications and just assume the failed one is the head of queue?
2. Create a handler per notification.

Of these, 1 looks decidedly dodgy and 2 is unscalable.

What's the recommended way of resending `PushNotification`s that fail due to an `ClientNotConnectedException`.

It would be nice if either `ApnsClient` or the `ClientNotConnectedException` itself exposed the `PushNotification`.

jchambers:
This closes #439 by introducing an application-specific `PushNotificationFuture`. The idea is that the push notification will always be associated with the future, so callers can get the original push notification even if the future fails. I'm opening this pull request to gather feedback on the approach.

One thing I like about this is that it's very application-specific and simplifies some internal plumbing a little (goodbye, `PushNotificationAndResponsePromise`). The main thing I dislike about this specific approach, though, is that we wind up with `SendPushNotificationFuture<T>`, but then `sendPushNotificationFuture.get()` returns something other than `T`. That feels confusing, and I'm concerned it'd be hard to communicate to users how to get details of the response.

Another approach I tried was to have `PushNotificationFuture<P, V> implements Future<V>`. That approach is basically fine; it makes the return type of the future clearer, but feels a lot more verbose and puts us in a position of maybe over-generalizing an application-specific concept.

What do you all think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,8 @@ final SimpleApnsPushNotification pushNotification;
 The process of sending a push notification is asynchronous; although the process of sending a notification and getting a reply from the server may take some time, the client will return a [`io.netty.util.concurrent.Future`](http://netty.io/4.1/api/io/netty/util/concurrent/Future.html) right away. You can use that `Future` to track the progress and eventual outcome of the sending operation. Note that an `io.netty.util.concurrent.Future` is an extension of the Java [`Future`](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html) interface that allows callers to add listeners and adds methods for checking the status of the `Future`.
 
 ```java
-final Future<PushNotificationResponse<SimpleApnsPushNotification>> sendNotificationFuture =
-        apnsClient.sendNotification(pushNotification);
+final PushNotificationFuture<SimpleApnsPushNotification, PushNotificationResponse<SimpleApnsPushNotification>>
+    sendNotificationFuture = apnsClient.sendNotification(pushNotification);
 ```
 
 The `Future` will complete in one of three circumstances:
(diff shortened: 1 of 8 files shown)


jchambers:
I brought back the two-type model (`PushNotificationFuture<P, V> implements Future<V>`) for review and discussion in de732ee.

jchambers:
…and @philipwhiuk, this might be especially interesting to you.

jchambers:
The more I'm thinking about it, the more I'm liking the two-type approach. It feels like a clearer extension of the existing `Future<V>` interface and demands less mind-bending from callers. It's verbose, but I think that's okay.

jchambers:
Smooshed everything down and also introduced a `PushNotificationResponseListener` for convenience. Will merge shortly.

PR #538, opened 2017-10-11
Create new Auth Token before expiration of current Auth Token

ksquaredkey:
Create a new Auth Token from the Auth key when the current Auth Token is about
to expire.  From the Apple documentation, an Auth Token is good for one hour.
If the current Auth Token is within 30 seconds of expiring or has already
expired, we create a new Auth Token.

The 30 seconds is an arbitrary fudge-factor to keep from hitting the race
condition where the token was not expired when we sent it but is expired
when the APNs server tries to process it.  It could most likely be much
less but there is no real cost to rotating the Auth Token 30 seconds early.


ksquaredkey:
@jchambers this was the token pre-creation idea I had.  Again, apologies if this is non-idiomatic Java.

ksquaredkey:
Oops, sorry, did not mean to close.  Have not done a pull request in a couple of years.

jchambers:
@ksquaredkey First and foremost, thank you very much for this contribution! I'm sorry for the very long delay in getting to it.

Given that Pushy automatically regenerates tokens when we get a "token expired" message from the APNs server and automatically re-sends notifications, can you elaborate on the case for why we'd want to preemptively expire tokens?

jchambers:
I'm going to close this for now pending more information. @ksquaredkey please give us a shout when you have a change to revisit this issue and we'll happily re-open it!

PR #539, opened 2017-10-11
More graceful netty shutdown

kurthong:
Dear Jon Chambers,

First of all, we've utilized this library into our projects and it works well and outperforms the others so far, Thank you for that.

My concern is the graceful shutdown. Considering Netty 4.0's nature and [doc](http://netty.io/wiki/user-guide-for-4.x.html#wiki-h3-16), in order to shutdown Netty gracefully, we just need to call `io.netty.util.concurrent.EventExecutorGroup.shutdownGracefully(long, long, TimeUnit)` method only. This method itself will close all channels related and also it provides a nice mechanism which is minimizing `submitted-but-not-processed tasks`(with the quiet period). But currently we close channels first and then shutdown event loop, so we don't utilize the mechanism. So, IMO, if we change the order of disposal, we can minimize `submitted-but-not-processed tasks`.

Thank you and cheers!

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java
@@ -290,23 +290,29 @@ public class ApnsClient {
             // event executor to notify listeners.
             final Promise<Void> closePromise = new DefaultPromise<>(GlobalEventExecutor.INSTANCE);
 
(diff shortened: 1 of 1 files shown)


kurthong:
I just thought that I might povide simply a idea but due to the introduction of connection pool, that became more complicated, so I will have a look futher and make another PR.

Thank you.

PR #558, opened 2017-11-26
[add]support retry mechanism when acquiring channel or sending notification fail

qiankunli:
add code to ApnsClient class:
1. add a queue to save the PushNotificationPromise if send it fail
2. add a period task to consume PushNotificationPromise in the queue

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java
@@ -98,7 +98,7 @@ public class ApnsClient<T extends ApnsPushNotification> {
     private static final IllegalStateException CLIENT_CLOSED_EXCEPTION =
             new IllegalStateException("Client has been closed and can no longer send push notifications.");
 
-    // can be a parameter of construct method
+    // can be a parameter of construct method;must set maxRetryCount to prevent from OOM
     private final int maxRetryCount = 3;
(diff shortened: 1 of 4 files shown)


jasonjoo2010:
it is suggested doing the retry in upper logic.
In best practice we will add semaphore like traffic quotation. Will it be invalid if we had many failed msgs in some condition?

qiankunli:
I just retry to send notification when acquiring channel fail or sending/writing fail which we can make sure that the notification must not be sent to Client. it is difficult to make the choice in upper logic.

I have tried to use CountDownLatch or guava-RateLimiter to realize Rate-Limiting, but there also is a lot of exception like Http2ChannelClosedException

qiankunli:
@jasonjoo2010 Can I add your WeChat, so that we can communicate more about APNs.

jasonjoo2010:
@qiankunli sure. jason_joo

I know what point you are meaning to now, but actually we rarely meet this now.

jchambers:
@qiankunli I'm EXTREMELY sorry to have let this sit for so long. Thank you very much for the contribution!

I definitely understand the need you're trying to address here, but I think this is complexity better left outside of Pushy (i.e. let callers decide how to handle this situation) for now.

Again, thank you very much for this thoughtful contribution, and I'm very sorry to have taken so long to respond.

PR #562, opened 2017-12-13
Add apns-id value to notification

aleksandr-lapushkin:
Providing an optional apns-id value for a notification provides great value for debugging any issues with APNs. Ideally, this value should also be filled out from the APNs response, but I fear that doing so may introduce performance problems. If you feel like this wouldn't be an issue, I'd be glad to introduce such parsing as well.


aleksandr-lapushkin:
Something went wrong, I'll re-create the PR.

aleksandr-lapushkin:
Wrong base chosen, correct PR here: https://github.com/relayrides/pushy/pull/563

PR #563, opened 2017-12-13
Add ability to specify 'apns-id' header value for a notification

aleksandr-lapushkin:
Providing an optional apns-id value for a notification provides great value for debugging any issues with APNs. Ideally, this value should also be filled out from the APNs response, but I fear that doing so may introduce performance problems. If you feel like this wouldn't be an issue, I'd be glad to introduce such parsing as well.


jchambers:
@aleksandr-lapushkin Thank you very much for the contribution!

I had a hard time assessing this without server-side support, and so I added that and adapted your changes in #577. I'm going to close this pull request for now since it's been absorbed into #577; let's move discussion there so we can do so with a more complete picture of the situation.

Cheers!

PR #571, opened 2018-01-03
Use an unresolved APNs hostname to improve GOAWAY handling

aleksandr-lapushkin:
As discussed in: [Connection Refused to APNs](https://groups.google.com/forum/#!topic/pushy-apns/5kpquH67JQU)

As a small update to the discussion: I did some extra debugging and was able to verify that upon re-establishing the connection to APNs after a GOAWAY is received the IP remains the same. This PR aims to fix that issue. 

I was also able to verify that with the fix in place, the IP indeed does change after we reconnect following a GOAWAY.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsClientBuilder.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsClientBuilder.java
@@ -160,7 +160,7 @@ public class ApnsClientBuilder {
      * @since 0.11
      */
     public ApnsClientBuilder setApnsServer(final String hostname, final int port) {
-        this.apnsServerAddress = new InetSocketAddress(hostname, port);
+        this.apnsServerAddress = InetSocketAddress.createUnresolved(hostname, port);
         return this;
     }
 


jmason:
makes sense to me, 👍

aleksandr-lapushkin:
Just as an update: we've been using our fork with this change for at least a week now in production. No connection errors of the kind that we observed earlier have appeared during this timeframe. Thus I'd say this was just what we needed.
Also, @jchambers, I'm guessing this will get merged by someone other than myself (for obvious reasons)?

jchambers:
> Also, @jchambers, I'm guessing this will get merged by someone other than myself (for obvious reasons)?

Yes. Putting out a new release is on my to-do list, but I still need to resolve some other (non-Pushy) priorities first.

aleksandr-lapushkin:
Got it. thanks.

jchambers:
@aleksandr-lapushkin we just released Pushy 0.12, which includes this fix. Thanks again for the contribution!

PR #576, opened 2018-01-18
Don't treat internal server errors as write exceptions

jchambers:
In days gone by, we wanted to treat upstream internal server errors as write exceptions instead of actual rejections. Two things have changed since we made that initial decision:

1. Apple has fixed an upstream bug that used to cause a lot of internal server exceptions when using token-based authentication, so these are now very rare.
2. Now that we have connection pooling, we have better means of addressing internal server errors (i.e. discarding unhealthy connections).

I propose removing that special-case handling and treating internal server errors just like any other rejection. The idea is that this would simplify things for end users by untangling which errors come from local/network problems and which errors come from the APNs server. What do you all think?

----

TODO:

- [x] Update docs


jmason:
That makes sense to me, 👍

PR #577, opened 2018-01-20
Add client and server support for `apns-id` headers

### LINKED ISSUE #559: support  apns-id
Hi,
First, thanks so much for the work you did.
To the issue: pls either allow adding the "apns-id" header sending, receiving or both.
There are (many?) cases where you get 200 OK and still the application does not show a push notification and this identifier could be used to further investigate what happened to the poor transaction.

According to "https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html" you could either send a UUID or get one in the response. if i had to choose only one i would take the APNS generated one but doing both may be the best.

This is rather small so I did take the lazy path asking you but if busy / too much trouble, I could do.

Shlomi

jchambers:
As requested in #559 and initially implemented by @aleksandr-lapushkin in #563, this adds support for the `apns-id` header in notifications sent to the server and in responses received from the server.

Having implemented this fully and benchmarked it, I recommend against moving forward because, in my assessment, the costs outweigh the benefits. As I understand the argument put forth in #559, the main benefit of exposing the APNs ID is that, in cases where the APNs server accepts a notification but the destination device does not report receiving the notification, callers could cite the APNs ID when contacting Apple to report the problem. I believe this is a flimsy argument because APNs itself is a best-effort protocol that _does not guarantee delivery_ even if the APNs server accepts a notification. As such, I believe the probability of Apple investigating and providing technical support for individual delivery attempts is identically zero.

As for the costs, I believe there are two major areas of concern:

1. Parsing UUIDs introduces a new Thing That Can Go Wrong™; it's possible that the UUID sent by the client could be malformed (or the server's expectation about format might change in the future), although I consider both of those cases very unlikely. The server could also send crazy data back (also unlikely) for its `apns-id` header, and we need to be prepared to handle UUID parsing failures in the client. In theory, the client could just report the `apns-id` from the server as a `String` to avoid parsing time and potential errors, and that may be worthy of consideration.
2. The much larger cost is, simply, performance (as @aleksandr-lapushkin predicted in #563). Adding proper handling significantly reduces throughput in benchmarks. Benchmark results are shown below ("scores" are the time taken to send 10,000 notifications, and so lower scores are better):

Ignoring `apns-id` (i.e. the current behavior in `master`):

| Concurrent connections | Score |   Error | Units |
|------------------------|-------|---------|-------|
|                      1 | 0.317 | ± 0.014 |  s/op |
|                      4 | 0.215 | ± 0.010 |  s/op |
|                      8 | 0.208 | ± 0.008 |  s/op |

With `apns-id` handling:

| Concurrent connections | Score |   Error | Units |
|------------------------|-------|---------|-------|
|                      1 | 0.422 | ± 0.021 |  s/op |
|                      4 | 0.249 | ± 0.017 |  s/op |
|                      8 | 0.226 | ± 0.007 |  s/op |

UUID parsing in Java is definitely [slower than it needs to be](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8192784), and so it's possible that we could use a more optimized implementation in Pushy. As mentioned before, we could also skip parsing for UUIDs included in responses from the server. Still, because I'm not persuaded by the existing arguments for including this feature at all, I'm not really inclined to go too crazy trying to make this happen.

I want to be very clear that I'm open to arguments in favor of this feature even though I haven't found the arguments so far persuasive. I'd like to leave this open for discussion, but would like to bring the discussion to a close within a couple weeks. Given a complete picture of the implementation and its implications, what do you all think?

This closes #559.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/server/MockApnsServerHandler.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/server/MockApnsServerHandler.java
@@ -52,8 +52,6 @@ class MockApnsServerHandler extends Http2ConnectionHandler implements Http2Frame
 
     private static final AsciiString APNS_ID_HEADER = new AsciiString("apns-id");
 
-    private static final Http2Headers SUCCESS_HEADERS = new DefaultHttp2Headers().status(HttpResponseStatus.OK.codeAsText());
-
     private static final int MAX_CONTENT_LENGTH = 4096;
 
     private static final Gson GSON = new GsonBuilder()
@@ -95,29 +93,13 @@ class MockApnsServerHandler extends Http2ConnectionHandler implements Http2Frame
         }
     }
 
-    private static class AcceptNotificationResponse {
-        private final int streamId;
-
-        AcceptNotificationResponse(final int streamId) {
-            this.streamId = streamId;
-        }
-
-        int getStreamId() {
-            return this.streamId;
-        }
-    }
-
-    private static class RejectNotificationResponse {
+    private static abstract class ApnsResponse {
         private final int streamId;
         private final UUID apnsId;
-        private final RejectionReason errorReason;
-        private final Date timestamp;
 
-        RejectNotificationResponse(final int streamId, final UUID apnsId, final RejectionReason errorReason, final Date timestamp) {
+        private ApnsResponse(final int streamId, final UUID apnsId) {
             this.streamId = streamId;
             this.apnsId = apnsId;
-            this.errorReason = errorReason;
-            this.timestamp = timestamp;
         }
 
         int getStreamId() {
@@ -125,7 +107,25 @@ class MockApnsServerHandler extends Http2ConnectionHandler implements Http2Frame
         }
 
         UUID getApnsId() {
-            return this.apnsId;
+            return apnsId;
+        }
+    }
+
+    private static class AcceptNotificationResponse extends ApnsResponse {
+        private AcceptNotificationResponse(final int streamId, final UUID apnsId) {
+            super(streamId, apnsId);
+        }
+    }
+
+    private static class RejectNotificationResponse extends ApnsResponse {
+        private final RejectionReason errorReason;
+        private final Date timestamp;
+
+        RejectNotificationResponse(final int streamId, final UUID apnsId, final RejectionReason errorReason, final Date timestamp) {
+            super(streamId, apnsId);
+
+            this.errorReason = errorReason;
+            this.timestamp = timestamp;
         }
 
         RejectionReason getErrorReason() {
@@ -137,15 +137,9 @@ class MockApnsServerHandler extends Http2ConnectionHandler implements Http2Frame
         }
     }
 
-    private static class InternalServerErrorResponse {
-        private final int streamId;
-
-        InternalServerErrorResponse(final int streamId) {
-            this.streamId = streamId;
-        }
-
-        int getStreamId() {
-            return this.streamId;
+    private static class InternalServerErrorResponse extends ApnsResponse {
+        private InternalServerErrorResponse(final int streamId, final UUID apnsId) {
+            super(streamId, apnsId);
         }
     }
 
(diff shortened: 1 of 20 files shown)


pushy/src/main/java/com/turo/pushy/apns/server/MockApnsServerHandler.java
```diff
@@ -294,19 +304,21 @@ private void handleEndOfStream(final ChannelHandlerContext context, final Http2S
     public void write(final ChannelHandlerContext context, final Object message, final ChannelPromise writePromise) {
         if (message instanceof AcceptNotificationResponse) {
             final AcceptNotificationResponse acceptNotificationResponse = (AcceptNotificationResponse) message;
-            this.encoder().writeHeaders(context, acceptNotificationResponse.getStreamId(), SUCCESS_HEADERS, 0, true, writePromise);
+
+            final Http2Headers headers = new DefaultHttp2Headers()
```
jchambers: Come to think of it, this is probably a major contributor to performance issues. The success headers used to be completely static, and now we're allocating new headers for each notification. We might be able to get back to a reuse-the-headers pattern, but I'd prefer to defer that until we've demonstrated a clear need to the `apns-id` header at all.

hzshlomi:
Well, for me it's kind of backwards - as flimsy as it gets I do have contacts I can use to cire the APNS ID, and since the server will generate a valid UUID for us given no ID of our own, and having that ID as a string is just fine for my purpose, I dont see the need for much debate over handing back a string generated by the server. That would cost zero performance, and should not break anything.
To sum up, i'm all for at least a getter for the APNS generated ID, as a string.

aleksandr-lapushkin:
Personally, having this feature is not something I'd consider critical, but it's just a pain to maintain our own fork and port this change to newer versions. However, since we expose individual deliveries to customers, it's good for providing general visibility, so that they can later investigate themselves, if needed. 

As for the implementation, I'd be totally fine without UUID parsing. All we need is some kind of identifier for the messages. A String will do just fine. 

I wonder if this is something that could be enabled by a flag during the client construction? That way this change wouldn't affect anyone who isn't interested in it.

jchambers:
@aleksandr-lapushkin In #563, the main thing you were focused on was sending an `apns-id` to the server. Is that something you still need to do, or would you also be happy to just receive the ID from the server?

Also, For Science™, I poked a bit to see if I could improve UUID parsing/stringifying and cooked up https://github.com/jchambers/fast-uuid-parser, which is about four times faster at parsing and six times faster at writing strings. Not sure how much that will help us here yet.

aleksandr-lapushkin:
Ideally, yes, I'd like to send it to APNs. To that end, the burden of UUID generation and validation can be entirely on the user of the library (crap in - crap out). However, I can still manage if I am only able to read the value.

jchambers:
I started to suspect that the thing that was actually getting measured by the benchmarks was mock server performance and not client performance. I created an entirely separate benchmark server in #585, and that does indeed reveal that the performance hit for the `apns-id` header is smaller than I initially believed. From there, I switched from the JVM-default `UUID#fromString(String)` and `UUID#toString()` to using [`fast-uuid`](https://github.com/jchambers/fast-uuid), and now the performance difference is basically undetectable.

I'm pretty comfortable moving forward with this if everybody agrees the implementation is reasonable.

aleksandr-lapushkin:
Having skimmed the changes, everything seems reasonable. At the very least on the API level.

hzshlomi:
I am very happy with everything that allows to fetch the apns-id out of the response from Apple actually, so I'm all for it! thanks a ton!

t1mek1ller:
Is APNs using apns-id to identify the push notification ?

jchambers:
> Is APNs server using apns-id to identify the push notification ?

Yes.

> if client do not set the apns-id, how does APNs server check whether it is the same request and not generate a new apns-id for the push notification ?

The APNs server does not perform any checks for duplicates.

> In order to avoid sending push notification duplicately, Is it necessary to set the apns-id in the client side ?

No. Again, this is not functionality provided by the APNs server.

PR #580, opened 2018-02-23
Update dependency versions in README

hupfis:


jchambers:
Hi!

Thanks kindly for the contribution, but I think I'm going to respectfully decline these specific changes. Since that dependency list is primarily for people managing things by hand, we want to list the lowest versions of the dependencies that we can possibly get away with. Newer ones will work—and we prefer them in many cases for bug fixes and performance enhancements—but those are the oldest ones we can get away with.

Cheers, and thank you again for the contribution!

Issue #583, opened 2018-03-07
Messages hangs after connection failure


anwerup:
When ApnsChannelFactory.create fails, for instance during a network outage, the caller only gets feedback on the listener of the first message and some random (assume it's timing base) number of subsequent messages. Most of the messages seem to be queued up an only processed at some later point when ApnsChannelFactory.create succeeds. Closing the ApnsClient seem to trigger failure callbacks on all of the outstanding messages.

I would expect to get a failure callback on the listeners of all messages sent during a network outage in a timely manner. The fact that they get queued up and don't seem to be processed until connections are reestablished creates three potential issues. 

- The queue size can get very large triggering OOM problems
- If reestablishing connection takes a long time messages can be out of date when they are delivered and make no sense to clients
- The lack of timely feedback on the failure by calling the listener makes it hard to catch and log failures in an appropriate manner

Is the behavior described in the first paragraph intended behavior or is this a bug?

Below is a partial test showing how to recreate the issue.

```java
@Test
public void sendMessagesSimple() throws InterruptedException, ExecutionException {

    SimpleApnsPushNotification pushNotification = createNotification(pushAddress, payload);

    for (int i = 0; i < 4; i++) {
        PushNotificationFuture<SimpleApnsPushNotification, PushNotificationResponse<SimpleApnsPushNotification>>
                pushNotificationFuture = apnsClient.sendNotification(pushNotification);

        pushNotificationFuture.addListener(
                (PushNotificationResponseListener<SimpleApnsPushNotification>) future -> {
                    System.out.println("Callback " + future);
                });
        Thread.sleep(1);
    }

    Thread.sleep(30000);
    System.out.println("\nWaited 30 sec. Shutting down\n");
    apnsClient.close();
}
```

This creates the following output when setting the APNS server name to an invalid server.

```
653  [main] INFO  com.turo.pushy.apns.SslUtil - Native SSL provider is available and supports ALPN; will use native provider. 
Callback handleConnectionCreationFailed
Callback handleWriteFailure
Callback PushNotificationPromise@7aa13b53(failure: io.netty.channel.ConnectTimeoutException: connection timed out: test.encap.no/85.221.67.15:9080)

Waited 30 sec. Shutting down

31051 [main] INFO  com.turo.pushy.apns.ApnsClient - Shutting down. 
Callback handleWriteFailure
Callback PushNotificationPromise@12a6f922(failure: java.lang.IllegalStateException: Channel pool has closed and no more channels may be acquired.)
[...]

PR #584, opened 2018-03-08
Fix inconsistent feedback on connection exception, Issue #583

anwerup:
See issue #583 'Messages hangs after connection failure'. This fix solves that issue and all other related issues where messages are queued indefinitely due to failure to acquire a connection.

Also added a test showing the problem. Running ApnsClientTest.testConsistentConnectionFailures without the fix will cause it to hang on line 672 until the test times out and fails.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -27,11 +27,11 @@
     <parent>
         <artifactId>pushy-parent</artifactId>
         <groupId>com.turo</groupId>
-        <version>0.12.0-1</version>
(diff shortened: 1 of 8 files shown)


jchambers:
As discussed in #583, I don't think this is a bug. What you're seeing is actually an intended effect of exponential back-off when connection attempts fail. Regardless of how we decide to proceed in #583, I'm afraid I have to decline this change because it's adding features to circumvent other features. If getting around exponential back-off is our goal, let's address exponential back-off directly.

Thank you very much for the contribution!

Issue #498, opened 2017-07-19
Publish Shaded JAR-File


lpfeup:
It would be useful to have a shaded jar with packaged dependencies and relocated to another package to prevent version conflicts (specially because of other dependencies which require a different version of netty).

jchambers:
I think this is a good idea, but I regret that it's highly unlikely to become a priority for us any time in the near term. Pull requests are welcome, though!

jchambers:
Lest it go unsaid/unnoticed, @ifrins opened a pull request to do this in #587, but we it might be a while before we can merge it.

Issue #540, opened 2017-10-11
Handle multiple sets of credentials with a single ApnsClient instance


jchambers:
Now that we've introduced connection pooling in #492, the next big step will be to get clients to choose connections from different sub-pools depending on the topic to which a notification should be sent (and therefore which credentials the client should use). To get there, we'll probably need to introduce some abstract notion of `ApnsCredentials` that has subclasses/implementations for X509 certificates and PKCS#8 signing keys, and then extend the existing connection pool infrastructure to add "keyed" pool functionality.

The end result will be that callers will only need to have a single `ApnsClient` instance for all of their destination apps, which will probably be a big win for high-volume users.

justinpermar:
+1 from a user who would benefit from this enhancement!

ChunBoL:
I also encountered the problem of handling multiple sets of documents, when can I support multiple sets of documents?

jchambers:
I'm starting in on some initial design work for this feature, and we're going to have to work around a few obstacles. The main ones right now are:

1. We have two possible types of credentials: X509 certificates for TLS-based authentication and EC private keys for token-based authentication.
2. Based on an X509 certificate's metadata, we can identify the APNs topics for which it can be used. We _can't_ tell which topics a signing key is good for, though, and have to rely on callers to tell us.

The general plan in my mind is to expand [`PooledObjectFactory`](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/PooledObjectFactory.java) (which creates connections to the APNs server) to be a `KeyedPooledObjectFactory` that would look something like this:

```java
interface KeyedPooledObjectFactory <K, T> {
    Future<T> create(K key, Promise<T> promise);
    Future<Void> destroy(K key, T object, Promise<Void> promise);
}
```

Elsewhere, we'd have an `ApnsCredentialProvider`:

```java
public interface ApnsCredentialProvider {
    Future<TlsApnsCredentials> getTlsCredentials(String topic);
    Future<ApnsSigningKey> getSigningKey(String topic);
}
```

I'm trying to decide what we should use for keys in the `KeyedPooledObjectFactory`. We could build everything around topics, which would be reasonably simple and straightforward, but maaaaay introduce some unnecessary overhead in cases where people are sending to lots of topics with the same credentials. If we keyed based on credentials instead, we'd have connections allocated per credential set rather than per topic.

In other words, it might be the difference between this:

> - com.example.app
>   - connection 1
>   - connection 2
[...]

PR #587, opened 2018-03-14
Use shading to build the packages

ifrins:
This PR adds generation of JARs shading the dependencies. It shades `io.netty`, `com.google.code.gson` and `org.apache.commons` into `com.turo.pushy.external`. 

This PR addresses #498. This is especially useful when used with newer versions of Spring Webflux, which use Netty with their Project Reactor, as it's more likely to produce version conflicts.

This also introduces the drawback of breaking source compatibility when using Netty extensions. For example, in my project, I had to rename the import of Netty's Future. 

What are your thoughts?

### DIFF
--- a/dropwizard-metrics-listener/pom.xml
+++ b/dropwizard-metrics-listener/pom.xml
@@ -40,6 +40,7 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>pushy</artifactId>
             <version>${project.version}</version>
+            <classifier>shaded</classifier>
         </dependency>
         <dependency>
             <groupId>io.dropwizard.metrics</groupId>
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -100,7 +100,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.2</version>
+                <version>3.1.0</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
@@ -109,6 +109,20 @@
                         </goals>
                         <configuration>
                             <finalName>${uberjar.name}</finalName>
+                            <relocations>
+                                <relocation>
+                                    <pattern>io.netty</pattern>
+                                    <shadedPattern>com.turo.pushy.external.io.netty</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.google.code.gson</pattern>
+                                    <shadedPattern>com.turo.pushy.external.com.google.code.gson</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>org.apache.commons</pattern>
+                                    <shadedPattern>com.turo.pushy.external.org.apache.commons</shadedPattern>
+                                </relocation>
+                            </relocations>
                             <transformers>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                                     <mainClass>org.openjdk.jmh.Main</mainClass>
--- a/pushy/pom.xml
+++ b/pushy/pom.xml
@@ -112,6 +112,37 @@
                 <plugins>
                     <!-- Download the alpn-boot.jar in advance to add it to the boot classpath. -->
                     <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-shade-plugin</artifactId>
+                        <version>3.1.0</version>
+                        <executions>
+                            <execution>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>shade</goal>
+                                </goals>
+                                <configuration>
+                                    <relocations>
+                                        <relocation>
+                                            <pattern>io.netty</pattern>
+                                            <shadedPattern>com.turo.pushy.external.io.netty</shadedPattern>
+                                        </relocation>
+                                        <relocation>
+                                            <pattern>com.google.code.gson</pattern>
+                                            <shadedPattern>com.turo.pushy.external.com.google.code.gson</shadedPattern>
+                                        </relocation>
+                                        <relocation>
+                                            <pattern>org.apache.commons</pattern>
+                                            <shadedPattern>com.turo.pushy.external.org.apache.commons</shadedPattern>
+                                        </relocation>
+                                    </relocations>
+                                    <shadedArtifactAttached>true</shadedArtifactAttached>
+                                    <shadedClassifierName>shaded</shadedClassifierName>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
                         <artifactId>maven-dependency-plugin</artifactId>
                         <executions>
                             <execution>


jchambers:
> This also introduces the drawback of breaking source compatibility when using Netty extensions. For example, in my project, I had to rename the import of Netty's Future.

I'm not 100% sure I follow; could you please elaborate?

jchambers:
…and, much more importantly, thank you for the contribution!

jchambers:
Interesting. Yeah, the other (worse?) case would be that something _else_ has Netty on the classpath, so the `import io.netty.util.concurrent.Future` thing would still work and risk mixing/matching versions in a way that's maybe hard to recognize or untangle.

> As a suggestion, if this is a concern, maybe there could be two pushy artifacts, the regular one (like the current) and a shaded one.

If we're going to publish shaded artifacts, I think that's the way to go. Poking quickly at Maven Central, I can find a number of projects that seem to follow that pattern. It's a brave new world for me, though; are you aware of any best practices in terms of publishing shaded/unshaded artifacts?

ifrins:
Right now, with the changes in this PR users should be able to decide if they want the shaded version by adding: `<classifier>shaded</classifier>` inside the `<dependency>` for pushy. I think this is pretty standard behaviour, for example, Spotify's [docker-client does it the same way](https://github.com/spotify/docker-client#a-note-on-shading).

I'm not sure if other changes would need to be done for the deployment of the release.

jchambers:
Ohhhhh yeah. Classifiers. Forgot about those ;)

Cool. I'll definitely have to do some prodding to get the deployment process right, but I think we're on the right track. Thanks kindly!

dropwizard-metrics-listener/pom.xml
```diff
@@ -40,6 +40,7 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>pushy</artifactId>
             <version>${project.version}</version>
+            <classifier>shaded</classifier>
```
jchambers: This feels a little odd to me; I don't think we want to make the Dropwizard metrics listener explicitly dependent upon the shaded version of Pushy. The [Maven docs](https://maven.apache.org/pom.html) aren't quite clear on this point; if we leave out a classifier here, will any classification of com.turo.pushy at the same version fulfill the requirement?

ifrins: Hmm. You're right, that's a bit strange. 

What could be done, for example, is also providing a shaded `dropwizard-metrics-listener` that had the shaded `pushy` dependency. However I don't even know if that's possible with maven.

What also could be done, which I think would be easier, is marking the `pushy` dependency in `dropwizard-metrics-listener` as optional, which would force users of the artifact, to also add `pushy` as a dependency explicitly (and then they'd be able to use either the shaded or the  non-shaded one).

justinpermar:
+1 on this PR. I was just about to start working on this one myself when I found this PR. So thank you for your efforts!

dropwizard-metrics-listener/pom.xml
justinpermar: From https://maven.apache.org/pom.html: ```Corresponds to the dependant artifact's packaging type. This defaults to jar```

justinpermar: @jchambers From : https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html```NOTE: In two of these dependency references, we had to specify the <type/> element. This is because the minimal set of information for matching a dependency reference against a dependencyManagement section is actually {groupId, artifactId, type, classifier}. In many cases, these dependencies will refer to jar artifacts with no classifier. This allows us to shorthand the identity set to {groupId, artifactId}, since the default for the type field is jar, and the default classifier is null.```

justinpermar:
is it important that netty-tcnative-boringssl-static dependency is not shaded? I'm unclear on how netty searched classpath at runtime to find this...

justinpermar:
As you guys already noted, with the shaded jar, Pushy exposes internal classes, thus breaking source compatibility for users. Consequently, it's possible that this PR needs to expand significantly and hide Netty dependencies from clients. Here is a sample error I get when dropping the shaded jar into my project: ```[ERROR] error: type mismatch;
[ERROR]  found   : io.netty.channel.nio.NioEventLoopGroup
[ERROR]  required: com.turo.pushy.external.io.netty.channel.EventLoopGroup
[ERROR]       .setEventLoopGroup(ActorManager.apnsNettyPool)```

justinpermar:
I believe a relocation is missed from the PR as-is. I also excluded various POM files from the dependencies that were packaged in the shaded jar:

```
                                    <relocations>
                                        <relocation>
                                            <pattern>io.netty</pattern>
                                            <shadedPattern>com.turo.pushy.external.io.netty</shadedPattern>
                                        </relocation>
                                        <relocation>
                                            <pattern>com.google.gson</pattern>
                                            <shadedPattern>com.turo.pushy.external.com.google.gson</shadedPattern>
                                        </relocation>
                                        <relocation>
                                            <pattern>com.google.code.gson</pattern>
                                            <shadedPattern>com.turo.pushy.external.com.google.code.gson</shadedPattern>
                                        </relocation>
                                        <relocation>
                                            <pattern>org.apache.commons</pattern>
                                            <shadedPattern>com.turo.pushy.external.org.apache.commons</shadedPattern>
                                        </relocation>
                                    </relocations>
                                    <transformers>
                                        <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
                                            <resources>
                                                <resource>META-INF/io.netty.versions.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-buffer/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-buffer/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec-http/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec-http/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec-http2/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec-http2/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-handler-proxy/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-handler-proxy/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-resolver/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-resolver/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec-socks/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-codec-socks/pom.properties</resource>
                                                <resource>META-INF/maven/com.google.code.gson/gson/pom.xml</resource>
                                                <resource>META-INF/maven/com.google.code.gson/gson/pom.properties</resource>
                                                <resource>META-INF/maven/commons-codec/commons-codec/pom.xml</resource>
                                                <resource>META-INF/maven/commons-codec/commons-codec/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-common/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-common/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-handler/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-handler/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-transport/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-transport/pom.xml</resource>
                                                <resource>META-INF/maven/io.netty/netty-tcnative-boringssl-static/pom.properties</resource>
                                                <resource>META-INF/maven/io.netty/netty-tcnative-boringssl-static/pom.xml</resource>
                                            </resources>
                                        </transformer>
                                    </transformers>
```

justinpermar:
Also, users will need to exclude the dependencies from the shaded jar, as the default POM is published to maven (and used by clients) (and the default POM lists the netty, etc. deps):
```
        <dependency>
            <groupId>com.turo</groupId>
            <artifactId>pushy</artifactId>
            <version>${pushy.version}</version>
            <classifier>shaded</classifier>
            <exclusions>
                <exclusion>
                    <groupId>io.netty</groupId>
                    <artifactId>*</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>com.google.code.gson</groupId>
                    <artifactId>gson</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>commons-codec</groupId>
                    <artifactId>commons-codec</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>org.slf4j</groupId>
                    <artifactId>slf4j-api</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

justinpermar:
FWIW, I have a project that uses Pushy and Twitter Finagle (which also uses Netty). I am able to compile my project and run everything successfully with the shaded jar (using my versions/changes I put in comments above).

jchambers:
> …I think the only way to do this properly is to update Pushy API to completely hide Netty.

I'm coming to the same conclusion, but am worried that the cure is worse than the disease. The two big things that we expose publicly (and deliberately) are:

1. [`io.netty.util.concurrent.Future`](http://netty.io/4.1/api/io/netty/util/concurrent/Future.html), which notably allows users to attach listeners to futures. Deriving from that is the set of `...FutureListener` interfaces. My enthusiasm for reinventing that set of wheels is a little low.
2. [`io.netty.channel.EventLoopGroup`](http://netty.io/4.1/api/io/netty/channel/EventLoopGroup.html), which I recognize is simultaneously most useful and most problematic for users combining Pushy with other Netty-based things. It's also important for users with lots of clients in play at the same time.

I guess we _can_ resolve the first problem as things stand now, though I'm not entirely convinced the benefit is worth the cost. To address the many-clients part of the second problem, I think we'd need to get #540 done so callers would still have a clear mechanism for controlling overall thread usage.

In all, I recognize that this is valuable for many users, but I think the costs are high and there are some big tasks to finish before we can seriously consider merging it. I say this not to shut down the effort, but to be clear that it will probably take significant time and effort before we can merge it.

Seem legit? Am I missing anything?

jchambers:
Some new upstream developments on the shading front:

- https://github.com/netty/netty-tcnative/pull/382/
- https://github.com/netty/netty/issues/7272
- https://github.com/netty/netty/pull/6995

jchambers:
> `io.netty.util.concurrent.Future`, which notably allows users to attach listeners to futures. Deriving from that is the set of ...FutureListener interfaces. My enthusiasm for reinventing that set of wheels is a little low.

Now that we've moved to Java 8 in #746, I think we have a reasonable path to getting around this obstacle. We might be able to use [`CompletableFuture`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) for all public methods that currently return an `io.netty.util.concurrent.Future`. It offers all of the same affordances, but is more Java-y. I think moving to `CompletableFuture` is probably a good move regardless, but did want to mention it here since I think it would knock down one of the two remaining obstacles to shading.

jchambers:
> We might be able to use `CompletableFuture` for all public methods…

This is covered in #757.

PR #594, opened 2018-03-31
Round-robin DNS resolution

jchambers:
This change introduces a round-robin DNS resolver. This will allow clients to spread their concurrent connections across a range of APNs and should, as a side effect of using Netty's DNS resolver, notice changes in the server list more quickly.

It has no effect if the client is using a proxy.


pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
```diff
@@ -84,7 +85,8 @@
         this.bootstrapTemplate.option(ChannelOption.TCP_NODELAY, true);
         this.bootstrapTemplate.remoteAddress(apnsServerAddress);
         this.bootstrapTemplate.resolver(proxyHandlerFactory == null ?
-                DefaultAddressResolverGroup.INSTANCE : NoopAddressResolverGroup.INSTANCE);
+                new RoundRobinDnsAddressResolverGroup(ClientChannelClassUtil.getDatagramChannelClass(eventLoopGroup),
```
jchambers: For readers/reviewers, this is the key change. Pretty much everything else is just adding dependencies on Netty's DNS resolver module or expanding `ClientChannelClassUtil` to cover datagram sockets, too.

palex-fpt:
This change leads to UDP socket leaks. Every instance of AddressResolverGroup subscribes to EventLoopGroup#terminationFuture. When EventLoopGroup is shared between several ApnsClient instances there is no way correctly terminate that resolvers as ApnsClient#close does not remove instance of resolver.

jchambers:
@palex-fpt Good catch. Could you please file a separate issue for this?

Issue #425, opened 2017-02-18
netty out of direct memory


ioilala:
my pushy version:0.8.1, jdk 1.8.0_66,netty 4.1.5.Final
I get such exception as follows after I use pushy to send IOS push notifycations for several months in my production environment. We send about 30,000,000 notifications every time, and several times per week.

> io.netty.util.internal.outofdirectmemoryerror: failed to allocate 16777216 byte(s) of direct memory (used: 17163092255, max: 17179869184),io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 16777216 byte(s) of direct memory (used: 17163092255, max: 17179869184),failed to allocate 16777216 byte(s) of direct memory (used: 17163092255, max: 17179869184),

Do I have to upgrade to the newest version of pushy(0.9.2) to solve this problem?

jchambers:
This is the first I've heard of the problem, and am not certain that upgrading to 0.9.2 would necessarily help.

To check the math, it looks like you're exhausting about 16Gb of direct memory after "several months." At 30,000,000 notifications per patch, an estimated 3 batches per week, four weeks per month, and (let's guess) 3 months until memory exhaustion, it sounds like you're sending somewhere around a billion notifications before you run out of memory. In other words, we're losing something like 16 bytes per notification.

I'm not entirely sure what could be going wrong here, but will see if I can reproduce the problem locally. To be clear, are you using `alpn-boot`, or are you using `netty-tcnative`?

kadar2012:
my pushy version is 0.9.1, jdk 1.8 ,netty 4.1.8.Final,
i am getting below issue, i can only able to send 2028 count of successful IOS push notification. after that i am getting below error,
using alpn-boot-8.1.10.v20161026-2.jar i am running this in local.

can sugges me how to solve this.

[23 Feb 2017 17:29:42]  WARN  io.netty.channel.AbstractChannel                   - Force-closing a channel whose registration task was not accepted by an event loop: [id: 0x00afe7a9]
java.lang.OutOfMemoryError: unable to create new native thread
	at java.lang.Thread.start0(Native Method)
	at java.lang.Thread.start(Thread.java:714)
	at io.netty.util.concurrent.ThreadPerTaskExecutor.execute(ThreadPerTaskExecutor.java:33)
	at io.netty.util.concurrent.SingleThreadEventExecutor.doStartThread(SingleThreadEventExecutor.java:847)
[...]

PR #598, opened 2018-04-09
Drop ALPN in favor of direct protocol negotiation

jchambers:
As [discussed on the mailing list](https://groups.google.com/d/msg/pushy-apns/P-RytjMRVYE/sZA2OkmlAwAJ), it turns out that we can just do direct protocol negotiation instead of using ALPN. This is ironic/semi-funny given how much woe and heartache ALPN has caused us and our users thus far, although the situation HAS been much better since we started relying directly on netty-tcnative.

Still, this is some baggage I'm happy to part with.


flozano:
Does this mean that we can drop netty-tcnative dependency? also, from the commit, it'd be great if we could configure it with java system properties instead (in addition?) of env variables.

jchambers:
> Does this mean that we can drop netty-tcnative dependency?

I would say "mostly yes," but I'd like to ease into that. After this, I think we can mostly view native SSL providers as optional (but still very helpful!), which would make it easier for end users to pick and choose between, say, netty-tcnative and Conscrypt. Also, the ciphers we need for HTTP/2 weren't widely added to the JDK until Java 8, so I'd say a native provider is still required-ish for Java 7, but not at all for Java 8.

> it'd be great if we could configure it with java system properties instead (in addition?) of env variables.

Interesting. Can you tell me more about the use case? The environment variable thing is largely a hack to support testing in CI environments, and I didn't imagine people would need it in the real world. I think this would only really be useful in cases where you have a native provider on the classpath, but don't want to use it for some reason. I'm not sure (a) why folks wouldn't want to use the native provider or (b) how often folks wind up in a situation where they want the JDK provider, but can't remove the native provider from the classpath. That doesn't mean the use case isn't legit, to be clear! It just means it's something I'll need help to understand.

Thanks!

flozano:
Well, basically it would be an attempt to avoid the TLS reconnection memory leak that has been reported in https://github.com/relayrides/pushy/issues/425.

jchambers:
Hm. Unless you know something I don't, I'm not sure that we've conclusively shown that #425 is a reconnection issue (my kingdom for enough consecutive hours to set up a real experiment and debug that thoroughly). (EDIT: just saw your edit).

Either way, though, couldn't you just exclude the netty-tcnative dependency?

jchambers:
@flozano I'm going to go ahead and merge this so we can start testing with our own systems before releasing publicly. If you still need more tools for selecting an SSL provider at runtime, could you please open a new issue?

Thanks kindly!

PR #599, opened 2018-04-11
Drop commons-codec as a dependency

jchambers:
[Apache Commons Codec](https://commons.apache.org/proper/commons-codec/) is a fine library that we were using for Base64 encoding/decoding for APNs signing/verification keys and authentication tokens. Even though it's a little more verbose, we can do the same thing with Netty's Base64 tools. Since we already depend on Netty, we can use that instead and have one less dependency to keep track of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@ If you don't use Maven (or something else that understands Maven dependencies, l
 
 - [netty 4.1.23](http://netty.io/)
 - [netty-tcnative-2.0.8.Final](http://netty.io/wiki/forked-tomcat-native.html)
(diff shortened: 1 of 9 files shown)


panchenko:
Another option would be using base64 from java8.

panchenko:
Could be just a matter of optimizing efforts.
as java 8 is 4 years old, I would expect that number of systems running java 7 _and_ having some development happening on that system is quite low already.

jchambers:
I'm happy to discuss JVM support, but I think it probably makes sense to open a separate issue or [mailing list](https://groups.google.com/d/forum/pushy-apns) thread. Seem legit?

PR #610, opened 2018-05-14
Update to Netty 4.1.25

jchambers:
This moves us up to the Netty 4.1.25, which was just released. @flozano you may be interested in https://github.com/netty/netty/pull/7903, which introduces a `-Dio.netty.handler.ssl.noOpenSsl=true/false` option for ignoring native SSL providers even if they're on the classpath.

TODO:

- [ ] ~Simplify SSL provider selection by relying on Netty defaults~ (won't do—the default native provider isn't reference-counted)
- [x] Switch unit tests to use `io.netty.handler.ssl.noOpenSsl` option

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ matrix:
 
         - os: linux
           jdk: oraclejdk8
-          env: PUSHY_SSL_PROVIDER=netty-tcnative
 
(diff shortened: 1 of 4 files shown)


flozano:
Thanks for noticing @jchambers !

PR #633, opened 2018-08-16
Close resolver groups when closing clients/channel factories

### LINKED ISSUE #632: UDP port leak on ApnsClient#close
**Problem:**
In #594 was introduced Round-robin DNS resolution. ApnsClient creates one instance of RoundRobinDnsAddressResolverGroup per client. Resolver subscribes to EventLoopGroup#terminationFuture for removal. When EventLoopGroup is created outside of ApnsClient that terminationFuture is never resolved. That leaves resolver (with open UDP port) leaked.
**Possible solution**
ApnsClient should close resolver at ApnsClient#close or allow to pass resolver instance through ApnsClientBuilder

jchambers:
This fixes #632 by closing address resolver groups when closing `ApnsChannelFactory` instances (which are in turn closed when we close an `ApnsClient` instance).

@palex-fpt look good to you?

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
@@ -109,7 +109,7 @@ class ApnsChannelFactory implements PooledObjectFactory<Channel>, Closeable {
         this.bootstrapTemplate.handler(new ChannelInitializer<SocketChannel>() {
 
             @Override
-            protected void initChannel(final SocketChannel channel) throws Exception {
+            protected void initChannel(final SocketChannel channel) {
                 final ChannelPipeline pipeline = channel.pipeline();
 
                 if (proxyHandlerFactory != null) {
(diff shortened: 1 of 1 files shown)


palex-fpt:
I have several Apple signing keys and use pool (based on org.apache.commons.pool2.KeyedPooledObjectFactory) of ApnsClient instances with shared EventLoopGroup. That instances are created/disposed based on application load. It looks like a bit overhead to create separate DnsNameResolver with its own name cache for each ApnsClient instance (hey, it resolves only 'api.push.apple.com'). 
It would be nice to have ability to pass single AddressResolverGroup to ApnsClientBuilder. but current solution is better than nothing.

jchambers:
> It would be nice to have ability to pass single AddressResolverGroup to ApnsClientBuilder.

It crossed my mind, but I don't think we'll do it for a few reasons:

1. The overhead cost, as I understand it, is low. It's true that we're creating one cache per `ApnsClient`, but I think the overhead there is fairly modest. This stands in contrast to event loop groups, for example, where an inability to reuse event loop groups would mean that we'd need to create at least a new thread (and with it a large stack) for every client.
2. The complexity is fairly high. We'd be creating opportunities for users to set an `AddressResolverGroup` that doesn't make sense given their proxy configuration (for example) and generally creating surface area for things to go wrong.
3. I think we have a better way to address the root cause, which it to let clients handle multiple sets of credentials (see #540). Once that's in place, I think most (all?) use cases for creating multiple clients will go away entirely.

Thanks kindly for the feedback! As you say, let's move forward with what's here now. If we need to adjust things in the future, we will, but I think the best move is to rely on multi-credential clients to eliminate the problem for us.

PR #634, opened 2018-08-16
Log rejected causes when trying to mark promises as failed

### LINKED ISSUE #631: Better log errors that should be reported by `tryFailure` but aren't
Explicitly I seem to be on a network where the connection-close timeout beats the dns resolution timeout but both happen.

This causes the generic `java.lang.IllegalStateException: Channel closed before HTTP/2 preface completed.` to beat the much more specific `java.net.UnknownHostException: failed to resolve 'api.development.push.apple.com' after 4 queries` in `com.turo.pushy.apns.ApnsChannelFactory#create`. I had to debug the code with a breakpoint there to find the cause of the issue.

Would it be possible to change anywhere that `promise.tryFailure(ex)` is called it can be changed to
```java
if (!promise.tryFailure(ex)) {
  logSwallowedError(ex);
}
```

jchambers:
This closes #631 by logging promise failure causes in cases where we try to mark the promise as failed, but the promise was already done for some other reason.

@mattnathan is this what you had in mind?

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
@@ -41,6 +41,8 @@ import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.concurrent.PromiseNotifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.Closeable;
 import java.net.InetSocketAddress;
@@ -69,6 +71,8 @@ class ApnsChannelFactory implements PooledObjectFactory<Channel>, Closeable {
     private static final AttributeKey<Promise<Channel>> CHANNEL_READY_PROMISE_ATTRIBUTE_KEY =
             AttributeKey.valueOf(ApnsChannelFactory.class, "channelReadyPromise");
 
+    private static final Logger log = LoggerFactory.getLogger(ApnsChannelFactory.class);
+
     @ChannelHandler.Sharable
     private static class ConnectionNegotiationErrorHandler extends ChannelHandlerAdapter {
 
@@ -76,7 +80,7 @@ class ApnsChannelFactory implements PooledObjectFactory<Channel>, Closeable {
 
         @Override
         public void exceptionCaught(final ChannelHandlerContext context, final Throwable cause) {
-            context.channel().attr(CHANNEL_READY_PROMISE_ATTRIBUTE_KEY).get().tryFailure(cause);
+            tryFailureAndLogRejectedCause(context.channel().attr(CHANNEL_READY_PROMISE_ATTRIBUTE_KEY).get(), cause);
         }
     }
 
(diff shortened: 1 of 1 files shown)


pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
```diff
@@ -249,4 +253,10 @@ public void close() {
             }
         }
     }
+
+    private static void tryFailureAndLogRejectedCause(final Promise<?> promise, final Throwable cause) {
+        if (!promise.tryFailure(cause)) {
+            log.debug("Could not set failure cause for promise because promise was already done. Rejected cause was {}", cause, cause);
```
jchambers: This logs both a summary of the exception AND a stack trace, which might be overkill. Very open to suggestions here.

jchambers:
Also worth pointing out: I narrowly scoped this to `ApnsChannelFactory`, though it's arguably useful in `ApnsChannelPool`, `ApnsClientHandler`, and `ApnsClient`, too. Having reviewed those cases, I think we'll get the highest benefit-to-noise ratio by focusing on `ApnsChannelFactory`, but am happy to be convinced otherwise!

jchambers:
…I also delayed when we try the vague "connection closed" failure reason in 788c841. That doesn't change any of the fundamentals of the issue, but should at least increase the probability that we'll have a more specific error message to offer.

jchambers:
Hm. Now I'm actually thinking that 788c841 is just asking for trouble and am inclined to revert it.

jchambers:
I had another idea, but I don't think it'll work. We _could_ use [`Throwable#addSuppressed(Throwable)`](https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#addSuppressed(java.lang.Throwable)) to add the "bounced" exception as a suppressed exception to whatever we get from `Promise#cause()`. The trouble is that it's _still_ a race condition. Callers might react to the failed `Promise` before we add the suppressed exception, and would never see that anything was added after the fact.

Sadly, it seems like logging is still the best option we have.

mattnathan:
Sorry, have been on annual leave and haven't checked email/github. The change looks good. 

I might say that a Warning rather than Debug log would be better but I leave that decision up to you :)

jchambers:
No worries! Thanks for getting back to me.

I did go back and forth on the logging level, but in keeping with the [logging policy in the README](https://github.com/relayrides/pushy#logging), I didn't feel like this rose to the level of a "serious error." I could definitely be persuaded, though!

mattnathan:
In my case it would fit into the "errors that may indicate a bug in caller's code" - though more bug in the callers system, i.e. bad dns setup. 

I guess the reason it's not sitting right with me is that without knowledge of how the pushy code works you wouldn't know that a 'closed early' type error could have more information in the debug log relating to a suppressed exception, causing undue headaches for the dev until they finally get around to enabling debug logging, probably after checking the source code like I did.

Checking Guava (because I tend to think it's a well designed lib) I notice that anywhere exceptions are suppressed it's logged as a warning.

jchambers:
Yeah—that makes sense to me. I'll bump it up to `warn`.

PR #635, opened 2018-08-21
Update to Netty 4.1.28

jchambers:
This brings us up to the latest version of Netty to take advantage of the latest bug fixes and performance enhancements. It sounds like there's an opportunity to improve performance when using `SslProvider.OPENSSL_REFCNT` (see the notes at http://netty.io/news/2018/07/10/4-1-26-Final.html), but I think we can defer that to a separate effort.

~I also want to look more into https://github.com/netty/netty/pull/8021, which sounds like it might improve (or fundamentally resolve?) the issue raised in #631.~ (EDIT: it doesn't)


jchambers:
> It sounds like there's an opportunity to improve performance when using SslProvider.OPENSSL_REFCNT

It seems like the benefits we'd see are mostly around establishing new connections, which I think is a small minority of where most users spend time right now. I think it's safe to call that a low-priority issue.

PR #641, opened 2018-09-11
Always try to fill the channel pool before reusing idle connections

### LINKED ISSUE #639: Channel pools not guaranteed to ever create more than one channel
```
        if (!this.isClosed) {
            final Channel channelFromIdlePool = ApnsChannelPool.this.idleChannels.poll();

            if (channelFromIdlePool != null) {
                if (channelFromIdlePool.isActive()) {
                    acquirePromise.trySuccess(channelFromIdlePool);
                } else {
                    // The channel from the idle pool isn't usable; discard it and create a new one instead
                    this.discardChannel(channelFromIdlePool);
                    this.acquireWithinEventExecutor(acquirePromise);
                }
            } else {
             // We don't have any connections ready to go; create a new one if possible.
            }
        }
```
I don't think it is a good idea. When I use it to push, I find some time channel.writeAndFlush use little time, so ApnsClient use only one channel to push. But the time of  apple server response is at 200 pieces per second and I have to set the semaphore to prevent the OOM. Therefore, ApnsClient just sends 200 pieces per second. So In my opinion, we should create max number of  channels in the initialization of ApnsChannelPool.

jchambers:
This changes the behavior of the `ApnsChannelPool` to always fill the pool (i.e. create up to `capacity` channels) before starting to reuse idle connections. As described in #639, this addresses an issue where there was no guarantee that a caller could ever send notifications fast enough to avoid just using the same channel over and over, which could limit throughput in locations with a slow connection to the APNs server.

This closes #639. @lianyuzhe look good to you?

Note: these changes are much easier to read with [whitespace-only changes hidden](https://github.com/relayrides/pushy/pull/641/files?w=1).

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelPool.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelPool.java
@@ -144,48 +144,51 @@ class ApnsChannelPool {
         assert this.executor.inEventLoop();
 
         if (!this.isClosed) {
-            final Channel channelFromIdlePool = ApnsChannelPool.this.idleChannels.poll();
+            // We always want to open new channels if we have spare capacity. Once the pool is full, we'll start looking
+            // for idle, pre-existing channels.
+            if (this.allChannels.size() + this.pendingCreateChannelFutures.size() < this.capacity) {
+                final Future<Channel> createChannelFuture = this.channelFactory.create(executor.<Channel>newPromise());
+                this.pendingCreateChannelFutures.add(createChannelFuture);
 
-            if (channelFromIdlePool != null) {
-                if (channelFromIdlePool.isActive()) {
-                    acquirePromise.trySuccess(channelFromIdlePool);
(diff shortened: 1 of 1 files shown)


lianyuzhe:
Maybe your network environment is good, but because I'm in Beijing and the connection with APNS is not stable. Some time of TLS handshake will time out, you will always to fail to create connection. At the same time, Previously established connections may still can push notifications. In my option, you can set a number to count the number of notification that pushed between the two create connection. And when the number reaches a threshold, you can create a connection. And the threshold is dynamic adjustment, it depends on if it is succeed to create a connection.

jchambers:
I think I get your point (though I'm wondering if we can achieve a similar result with less complexity), but that sounds like something we can approach as a separate enhancement. I'm inclined to move forward with this to address the immediate issue and then see if more is actually needed in practice. Do you view the issue you've raised as a blocker, or do you agree that we can take that on as a separate effort?

PR #651, opened 2018-11-26
Added `summary-arg` and `summary-arg-count` to the payload builder.

mattozer:
This adds support for the summary-arg and  summary-arg-count fields to the payload builder.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
@@ -540,13 +540,15 @@ public class ApnsPayloadBuilder {
      *
      * <blockquote>The string the notification adds to the category’s summary format string.</blockquote>
      *
-     * <p>For example when defining an UNNotificationCategory, passing a format string like
-     * "%u more messages from %@" to the categorySummaryFormat argument, will produce
+     * <p>For example in iOS 12, when defining an {@code UNNotificationCategory}, passing a format string like
+     * "%u more messages from %@" to the {@code categorySummaryFormat} argument, will produce
      * "x more messages from {summaryArgument}.</p>
      *
-     * <p>By default, it says “x more notifications”, no summary argument is included.</p>
+     * <p>In iOS 12, the default summary format string in English is "%u more notifications" and does not have a
+     * placeholder for a summary argument. By default, no summary argument is included in push notification payloads.</p>
      *
-     * @param summaryArgument the summary argument for this notification
+     * @param summaryArgument the summary argument for this notification; if
+     * {@code null}, the {@code summary-arg} key is omitted from the payload entirely
      *
      * @return a reference to this payload builder
      *
@@ -566,13 +568,19 @@ public class ApnsPayloadBuilder {
      *
      * <blockquote>The number of items the notification adds to the category’s summary format string.</blockquote>
      *
-     * <p>For example when defining an UNNotificationCategory, passing a format string like
-     * like "%u more podcasts" to the categorySummaryFormat argument, will produce "{argumentSummaryCount} more podcasts"
+     * <p>By default, all notifications count as a single "item" in a group of notifications. The summary argument count
+     * controls how many "items" in a "stack" of notifications are represented by a specific notification.
+     * If, for example, a notification informed a user that seven new podcasts are available, it might be helpful to set
+     * the summary argument count to 7. When "stacked," the notification would contribute an item count of 7
+     * to the total number of notifications reported in the summary string (for example, "7 more podcasts").
      * </p>
      *
-     * <p>By default, summary argument count is 1 and it can not be 0.</p>
+     * <p>By default, notifications count as a single "item" in a group of notifications, and so the default
+     * summary argument count is 1 (even if no summary argument count is specified in the notification payload).
+     * If specified, summary argument count must be positive.</p>
      *
      * @param summaryArgumentCount the summary argument count for this notification
+     * {@code null}, the {@code summary-arg-count} key is omitted from the payload entirely
      *
      * @return a reference to this payload builder
      *
@@ -583,6 +591,9 @@ public class ApnsPayloadBuilder {
      * @since 0.13.6
      */
     public ApnsPayloadBuilder setSummaryArgumentCount(final Integer summaryArgumentCount) {
+        if (summaryArgumentCount != null && summaryArgumentCount < 1) {
+            throw new IllegalArgumentException("Summary argument count must be positive.");
+        }
         this.summaryArgumentCount = summaryArgumentCount;
         return this;
     }
--- a/pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
@@ -85,8 +85,8 @@ public class ApnsPayloadBuilder {
     private static final String CONTENT_AVAILABLE_KEY = "content-available";
     private static final String MUTABLE_CONTENT_KEY = "mutable-content";
     private static final String THREAD_ID_KEY = "thread-id";
-    private static final String SUMMARY_ARGUMENT = "summary-arg";
-    private static final String SUMMARY_ARGUMENT_COUNT = "summary-arg-count";
+    private static final String SUMMARY_ARGUMENT_KEY = "summary-arg";
+    private static final String SUMMARY_ARGUMENT_COUNT_KEY = "summary-arg-count";
     private static final String URL_ARGS_KEY = "url-args";
 
     private static final String ALERT_TITLE_KEY = "title";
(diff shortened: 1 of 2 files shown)


pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,47 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
         return this;
     }
 
+    /**
+     * <p>Sets the summary argument for this notification. The summary argument is:</p>
+     *
+     * <blockquote>The argument to be inserted in the summary for this notification when the system groups the
```
jchambers: If this is a quote, can we cite the source (perhaps in a `@see` tag)? If it's not a quote, I think we can just use a normal paragraph (`<p>...</p>`).

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -671,6 +717,14 @@ public String buildWithMaximumLength(final int maximumPayloadSize) {
                 aps.put(THREAD_ID_KEY, this.threadId);
             }
 
+            if (this.summaryArgument != null) {
+                aps.put(SUMMARY_ARGUMENT, this.summaryArgument);
```
jchambers: Do we want these in the `aps` dictionary, or in the `aps.alert` dictionary?

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -82,6 +85,8 @@
     private static final String CONTENT_AVAILABLE_KEY = "content-available";
     private static final String MUTABLE_CONTENT_KEY = "mutable-content";
     private static final String THREAD_ID_KEY = "thread-id";
+    private static final String SUMMARY_ARGUMENT = "summary-arg";
```
jchambers: Trivial: can we append `_KEY` to these names for consistency?

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: Updated comments in ed6aed6.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: In `aps.alert`, moved them in ed6aed6.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: Added in ed6aed6.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,58 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
         return this;
     }
 
+    /**
+     * <p>Sets the summary argument for this notification. The summary argument is:</p>
+     *
+     * <blockquote>The string the notification adds to the category’s summary format string.</blockquote>
+     *
+     * <p>For example when defining an UNNotificationCategory, passing a format string like
```
jchambers: It's implied, but I might explicitly call out that the "summary format string" mentioned in the quote above is defined by the receiving app (as opposed to whoever is sending the push notification). Not all users of Pushy will be familiar with the details of the iOS SDK that actually handles notifications; with that in mind, taking some extra caution to explain when things are outside of Pushy's scope can be a big help.

Also, as a suuuuuper nitpick, could we do `{@code UNNotificationCategory}`? That will format the text as monospace and make it look more, y'know, code-y.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,58 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
...
+     * "%u more messages from %@" to the categorySummaryFormat argument, will produce
+     * "x more messages from {summaryArgument}.</p>
+     *
+     * <p>By default, it says “x more notifications”, no summary argument is included.</p>
```
jchambers: Minor phrasing suggestion:

> In iOS 12, the default summary format string in English is "%u more notifications" and does not have a placeholder for a summary argument. By default, no summary argument is included in push notification payloads.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,58 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
...
+     *
+     * @param summaryArgument the summary argument for this notification
```
jchambers: Can we call out what happens if we set this to `null`?

```
@param summaryArgument the summary argument for this notification; if
{@code null}, the {@code summary-arg} key is omitted from the payload entirely
```

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,58 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
...
+        this.summaryArgument = summaryArgument;
+        return this;
+    }
+
+    /**
+     * <p>Sets the summary argument count for this notification. The summary argument count is:</p>
+     *
+     * <blockquote>The number of items the notification adds to the category’s summary format string.</blockquote>
+     *
+     * <p>For example when defining an UNNotificationCategory, passing a format string like
```
jchambers: I think it might be helpful to be more verbose (maybe give a real-world example?), particularly given how long it took the two of us to figure out what was going on here this morning ;)

> By default, all notifications count as a single "item" in a group of notifications. The summary argument count controls how many "items" in a "stack" of notifications are represented by a specific notification. If, for example, a notification informed a user that seven new podcasts are available, it might be helpful to set the summary argument count to 7. When "stacked," the notification would contribute an item count of 7 to the total number of notifications reported in the summary string (for example, "7 more podcasts").

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,58 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
...
+     * like "%u more podcasts" to the categorySummaryFormat argument, will produce "{argumentSummaryCount} more podcasts"
+     * </p>
+     *
+     * <p>By default, summary argument count is 1 and it can not be 0.</p>
```
jchambers: Minor phrasing suggestion:

> By default, notifications count as a single "item" in a group of notifications, and so the default summary argument count is 1 (even if no summary argument count is specified in the notification payload). If specified, summary argument counts must be positive.

Relatedly, should we actually enforce that restriction in this setter? We could throw an `IllegalArgumentException` if the count is non-null and less than 1.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -530,6 +535,58 @@ public ApnsPayloadBuilder setThreadId(final String threadId) {
...
+     *
+     * @param summaryArgumentCount the summary argument count for this notification
```
jchambers: Can we specify what happens if this is `null`?

mattozer: Done in cd6eb42.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: Throwing the exception in cd6eb42.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: Changed in cd6eb42.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: Done in cd6eb42.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: Done in cd6eb42.

pushy/src/main/java/com/turo/pushy/apns/util/ApnsPayloadBuilder.java
mattozer: > mentioned in the quote above is defined by the receiving app

Makes sense. Done in cd6eb42. Let me know if you'd like to reword.

jchambers:
🤘 

Thanks, @mattozer!

PR #656, opened 2018-12-19
Including ALPN supported protocols in Mock Server

netofontes:
Fixing an issue in mock server where the ALPN extension fields for protocol negotiation were not being added to the 'Server Hello' TLS package.

When using mock server with different HTTP/2-compliant clients only accepting HTTP/2 protocol, connection couldn't be established as mock server wasn't setting ALPN extension fields with the supported protocols.

'_Server Hello_' package from the **real** APNS server.
image

'_Server Hello_' package from mock server **before** the fix.
image

'_Server Hello_' package from mock server **after** the fix.
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,6 @@ doc
 _site
 #ignore intellij files
 *.iml
+*.ipr
+*.iws
 .idea/
--- a/pushy/src/main/java/com/turo/pushy/apns/server/BaseHttp2ServerBuilder.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/server/BaseHttp2ServerBuilder.java
@@ -291,6 +291,15 @@ abstract class BaseHttp2ServerBuilder <T extends BaseHttp2Server> {
                 sslContextBuilder.trustManager(this.trustedClientCertificates);
             }
 
+            // Mock server needs to be able to inform to ALPN-enabled HTTP clients
+            // which protocols are supported during the protocol negotiation phase.
+            // In this case, mock server should only support HTTP/2 in order to mock real APNS servers.
+            sslContextBuilder.applicationProtocolConfig(new ApplicationProtocolConfig(
+                    ApplicationProtocolConfig.Protocol.ALPN,
+                    ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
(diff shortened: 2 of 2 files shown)


netofontes:
@jchambers, any chance of having this fix in a possible 0.13.7 release in the foreseeable future?

jchambers:
We deliberately removed ALPN support some time ago to remove dependencies on `alpn-agent` or `netty-tcnative` (see #598). Can you say more about your use case?

netofontes:
Without ALPN on the server-side, HTTP clients won't be able to negotiate and establish HTTP/2 connections using TLS, only HTTP/2 over cleartext TCP will be allowed. See [HTTP/2 spec](https://http2.github.io/http2-spec/#known-http).

Our use case requires a HTTP/2-complaint mock server so that we are able to test functionality and performance of our own APNS client. As we want the mock server to be close as possible of real APNS server, we need the mock to be able to handle HTTP/2 request with TLS to guarantee that our client fully complies with the HTTP/2 APNS API.

`netty-tcnative` or any other agent to enable ALPN in TLS will only be required as a run time dependency. If having them as run time dependency can be a problem, then we can add a configuration (perhaps a system property) to enable/disable the code snippet configuring the application protocol for the `sslContentBuilder`. So, anyone who wishes to use HTTP/2 with TLS can just enable that property and provide the necessary run time libraries in the classpath.

jchambers:
Ah—the key detail I was missing is that you're using this to test a client other than Pushy (which is a totally valid use case!).

Let me think on it some, though. Accurate emulation of the APNs server is, indeed, a design goal, but re-introducing ALPN also carries costs.

jchambers:
> have a `useAlpn` flag in the mock server builder.

…though I guess if we want to go down that road, we can just merge this pull request as-is and I can do a quick follow-up to add the option to `MockApnsServerBuilder`.

jchambers:
> we can just merge this pull request as-is and I can do a quick follow-up

Yeah. Let's do that.

jchambers:
…and I've followed up in #660.

netofontes:
Sorry about the delay. That’s perfect. 
When would be possible to have a new release from the master branch?
Thanks a lot!

PR #657, opened 2018-12-22
Consolidate flushes to reduce overhead

### LINKED ISSUE #655: Consider batching flushes for reduced overhead
We might be able to use Netty's [`FlushConsolidationHandler`](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/flush/FlushConsolidationHandler.java) to reduce the ratio of flushes to push notifications. We tried a DIY approach in #276/#278, but it wound up causing some mystery bugs (and let's be clear—those mystery bugs were certainly my fault). With better upstream support, though, it might be worth a second look.

jchambers:
This change batches calls to `flush` together where possible. `flush` is a relatively expensive call. By batching calls to `flush` together, we can sell out write latency (which isn't really a primary concern) to pick up some gains in overall throughput (which _is_ a primary concern). Some benchmarks with a local client and server show a significant improvement with this change in place:

### Before

| Concurrent connections | Time to send 10,000 notifications |
|------------------------|-----------------------------------|
|                      1 |             0.336 ± 0.003 seconds |
|                      4 |             0.231 ± 0.019 seconds |
|                      8 |             0.221 ± 0.013 seconds |

### After

| Concurrent connections | Time to send 10,000 notifications |
|------------------------|-----------------------------------|
|                      1 |             0.312 ± 0.002 seconds |
|                      4 |             0.160 ± 0.009 seconds |
|                      8 |             0.152 ± 0.013 seconds |

This closes #655.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
@@ -28,6 +28,7 @@ import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.*;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.handler.codec.http2.Http2FrameLogger;
+import io.netty.handler.flush.FlushConsolidationHandler;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.timeout.IdleStateHandler;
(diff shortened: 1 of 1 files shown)


jchambers:
Yow. Combined with #658, this makes a _huge_ difference in throughput for local benchmarks:

### Pushy v0.13.6

| Concurrent connections | Time to send 10,000 notifications |
|------------------------|-----------------------------------|
|                      1 |             0.333 ± 0.002 seconds |
|                      4 |             0.251 ± 0.010 seconds |
|                      8 |             0.212 ± 0.018 seconds |

### After updating to Netty 4.1.32 and consolidating flushes

| Concurrent connections | Time to send 10,000 notifications |
|------------------------|-----------------------------------|
|                      1 |             0.257 ± 0.002 seconds |
|                      4 |             0.137 ± 0.007 seconds |
|                      8 |             0.116 ± 0.006 seconds |

That's nearly double the throughput at higher concurrency levels.

normanmaurer:
Woohoo that’s awesome

PR #661, opened 2018-12-29
Use fancier server builder generics

jchambers:
This change modifies how we're using generics in mock server builders so we don't have to override everything in subclasses of `BaseHttp2ServerBuilder` for type safety. Now, this definitely comes with some tradeoffs (see https://stackoverflow.com/questions/38099005/how-to-return-this-in-the-generic-class for additional discussion). The main problem is that we're using an unsafe cast in `BaseHttp2ServerBuilder` (`return (T) this`), and careless subclassers could break things. That said, I think it's unlikely that folks are subclassing server builders and, if they are and do something weird that causes casting problems, _well maybe they should fix that._

I'm very open to feedback on these ideas, but I think the gains in simplicity and reduced repetition justify the costs (primarily risk of making a mistake in an improbable use case).

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/server/BaseHttp2ServerBuilder.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/server/BaseHttp2ServerBuilder.java
@@ -35,7 +35,7 @@ import java.io.InputStream;
 import java.security.PrivateKey;
 import java.security.cert.X509Certificate;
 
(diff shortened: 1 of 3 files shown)


jchambers:
…but then again, this also appears to be very confusing to `javadoc`. If I can't get that to work out as expected, then I'll back off on this idea and stick to the repetitive approach.

PR #689, opened 2019-04-02
Turn the Pushy JAR file into an OSGi bundle

arunan-interlink:
This is slightly complicated by needing to explicitly import Netty packages that are only referenced via Class.forName().

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,6 +159,13 @@
         <pluginManagement>
             <plugins>
                 <plugin>
+                    <groupId>org.apache.felix</groupId>
+                    <artifactId>maven-bundle-plugin</artifactId>
+                    <!-- 3.5.1 is the last version supporting Java 7. -->
+                    <version>3.5.1</version>
+                </plugin>
+
+                <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>3.7.0</version>
--- a/pushy/pom.xml
+++ b/pushy/pom.xml
@@ -26,6 +26,7 @@
 
     <artifactId>pushy</artifactId>
     <version>0.13.11-SNAPSHOT</version>
+    <packaging>bundle</packaging>
     <name>Pushy</name>
 
     <parent>
@@ -96,6 +97,23 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Import-Package>
+                            <!-- Classes from the following packages are only loaded using Class.forName(), so BND can not detect they are required imports. -->
+                            io.netty.channel.socket.nio,
+                            io.netty.channel.socket.oio,
+                            <!-- This tells BND to add all other required imports that it does detect, as normal. -->
+                            *,
+                        </Import-Package>
+                    </instructions>
+                </configuration>
+            </plugin>
+
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>


arunan-interlink:
I have run this in Apache Karaf and my use case appears to work. I tried running the tests and got failures, but they appeared to be certificate related so I might just be missing something from my key store.

arunan-interlink:
Repushed with an older version of the maven-bundle-plugin that uses and older BND that should support Java 7. Checked that the headers are still OK.

pushy/pom.xml
```diff
@@ -96,6 +97,23 @@
...
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Import-Package>
+                            <!-- Classes from the following packages are only loaded using Class.forName(), so BND can not detect they are required imports. -->
+                            io.netty.channel.socket.nio,
+                            io.netty.channel.socket.oio,
```
jchambers: Looking at this with fresh eyes, I can't imagine any universe in which somebody would use OIO on purpose. Maybe we should just drop support for OIO entirely. That's definitely a separate effort, though.

pom.xml
```diff
@@ -158,6 +158,12 @@
     <build>
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <groupId>org.apache.felix</groupId>
+                    <artifactId>maven-bundle-plugin</artifactId>
+                    <version>3.5.1</version>    <!-- The last version supporting Java 7. -->
```
jchambers: Trivial: would you mind putting the comment on the preceding line?

```suggestion
                    <!-- The last version supporting Java 7. -->
                    <version>3.5.1</version>    
```

arunan-interlink: Sure, repushed!

pushy/pom.xml
arunan-interlink: Yes; while it's in the code, it would be best to have it in thee OSGi headers.

arunan-interlink:
Thank you for looking at this so quickly. We have another PR on the way, just testing and tidying it up. It would be handy for that to be considered before your 0.14.0 release so I will get it pushed ASAP.

OSGi is many things, but in this case it may be best to see it as something like the Java platform module system that supports module versioning as well. It allows declaring what dependencies a JAR should have and enforcing that. It does this by adding metadata to JAR files that OSGi runtimes pick up.

Since OSGi systems work in a world where many if not most JAR files do not have OSGi metadata, they allow auto-generating this dependency information by analysing a JAR's byte code to see what packages outside of the JAR are referenced. In cases which do not use reflection or Class.forName() this created accurate metadata. Unfortunately, since Pushy uses Class.forName() in a few places, and the classes referenced are in packages that are not otherwise referenced in the byte code, the auto generation does not detect those packages are needed and then at runtime those classes are blocked from being loaded as their packages are not linked in.

The pushy-parent bundle remains a POM. The pushy module is currently using the default packaging of "jar". Changing it to "bundle" invokes the maven-bundle-plugin. This only affects the metadata in the generated JAR file, and the generated artifact is still a JAR file that works normally in non-OSGi environments. Building this branch and comparing the unzipped JAR file to the version built on master shows that only files in the "META-INF" directory are different - the diff is below.

`diff -r -u pushy-0.14.0-SNAPSHOT/META-INF/MANIFEST.MF pushy-0.14.0-SNAPSHOT-osgi/META-INF/MANIFEST.MF
--- pushy-0.14.0-SNAPSHOT/META-INF/MANIFEST.MF	2019-04-03 08:59:54.000000000 +0100
+++ pushy-0.14.0-SNAPSHOT-osgi/META-INF/MANIFEST.MF	2019-04-03 07:57:30.000000000 +0100
@@ -1,5 +1,44 @@
 Manifest-Version: 1.0
-Built-By: balasuba
-Created-By: Apache Maven 3.5.4
+Bnd-LastModified: 1554278251374
 Build-Jdk: 1.8.0_181
-
+Built-By: balasuba
+Bundle-Description: A Java library for sending push notifications
+Bundle-DocURL: https://turo.com/
+Bundle-License: http://opensource.org/licenses/MIT
+Bundle-ManifestVersion: 2
+Bundle-Name: Pushy
+Bundle-SymbolicName: com.turo.pushy
+Bundle-Vendor: Turo
+Bundle-Version: 0.14.0.SNAPSHOT
+Created-By: Apache Maven Bundle Plugin
+Export-Package: com.turo.pushy.apns;uses:="com.turo.pushy.apns.auth,co
+ m.turo.pushy.apns.proxy,com.turo.pushy.apns.util.concurrent,io.netty.
+ channel,io.netty.handler.codec.http2,io.netty.handler.ssl,io.netty.ut
+ il.concurrent,javax.net.ssl";version="0.14.0",com.turo.pushy.apns.aut
+ h;uses:="io.netty.util";version="0.14.0",com.turo.pushy.apns.proxy;us
+ es:="io.netty.handler.proxy";version="0.14.0",com.turo.pushy.apns.ser
+ ver;uses:="com.turo.pushy.apns,com.turo.pushy.apns.auth,io.netty.buff
+ er,io.netty.channel,io.netty.handler.codec.http2,io.netty.handler.ssl
+ ,io.netty.util.concurrent,javax.net.ssl";version="0.14.0",com.turo.pu
+ shy.apns.util;uses:="com.google.gson,com.turo.pushy.apns";version="0.
+ 14.0",com.turo.pushy.apns.util.concurrent;uses:="com.turo.pushy.apns,
+ io.netty.util.concurrent";version="0.14.0"
+Import-Package: com.eatthepath.uuid,com.google.gson,com.google.gson.an
+ notations,com.turo.pushy.apns,com.turo.pushy.apns.auth,com.turo.pushy
+ .apns.proxy,com.turo.pushy.apns.util,com.turo.pushy.apns.util.concurr
+ ent,io.netty.bootstrap;version="[4.1,5)",io.netty.buffer;version="[4.
+ 1,5)",io.netty.channel;version="[4.1,5)",io.netty.channel.group;versi
+ on="[4.1,5)",io.netty.channel.nio;version="[4.1,5)",io.netty.channel.
+ socket;version="[4.1,5)",io.netty.handler.codec;version="[4.1,5)",io.
+ netty.handler.codec.base64;version="[4.1,5)",io.netty.handler.codec.h
+ ttp;version="[4.1,5)",io.netty.handler.codec.http2;version="[4.1,5)",
+ io.netty.handler.flush;version="[4.1,5)",io.netty.handler.proxy;versi
+ on="[4.1,5)",io.netty.handler.ssl;version="[4.1,5)",io.netty.handler.
+ timeout;version="[4.1,5)",io.netty.resolver;version="[4.1,5)",io.nett
+ y.resolver.dns;version="[4.1,5)",io.netty.util;version="[4.1,5)",io.n
+ etty.util.collection;version="[4.1,5)",io.netty.util.concurrent;versi
+ on="[4.1,5)",javax.net.ssl,org.slf4j;version="[1.7,2)",io.netty.chann
+ el.socket.nio;version="[4.1,5)",io.netty.channel.socket.oio;version="
+ [4.1,5)"
+Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
+Tool: Bnd-3.5.0.201709291849
diff -r -u pushy-0.14.0-SNAPSHOT/META-INF/maven/com.turo/pushy/pom.properties pushy-0.14.0-SNAPSHOT-osgi/META-INF/maven/com.turo/pushy/pom.properties
--- pushy-0.14.0-SNAPSHOT/META-INF/maven/com.turo/pushy/pom.properties	2019-04-03 08:59:54.000000000 +0100
+++ pushy-0.14.0-SNAPSHOT-osgi/META-INF/maven/com.turo/pushy/pom.properties	2019-04-03 07:57:30.000000000 +0100
@@ -1,4 +1,5 @@
-#Created by Apache Maven 3.5.4
+#Generated by org.apache.felix.bundleplugin
+#Wed Apr 03 08:57:31 BST 2019
 version=0.14.0-SNAPSHOT
 groupId=com.turo
 artifactId=pushy
diff -r -u pushy-0.14.0-SNAPSHOT/META-INF/maven/com.turo/pushy/pom.xml pushy-0.14.0-SNAPSHOT-osgi/META-INF/maven/com.turo/pushy/pom.xml
--- pushy-0.14.0-SNAPSHOT/META-INF/maven/com.turo/pushy/pom.xml	2019-04-03 08:59:40.000000000 +0100
+++ pushy-0.14.0-SNAPSHOT-osgi/META-INF/maven/com.turo/pushy/pom.xml	2019-04-02 09:23:50.000000000 +0100
@@ -26,6 +26,7 @@
 
     <artifactId>pushy</artifactId>
     <version>0.14.0-SNAPSHOT</version>
+    <packaging>bundle</packaging>
     <name>Pushy</name>
 
     <parent>
@@ -97,6 +98,23 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Import-Package>
+                            
+                            io.netty.channel.socket.nio,
+                            io.netty.channel.socket.oio,
+                            
+                            *,
+                        </Import-Package>
+                    </instructions>
+                </configuration>
+            </plugin>
+
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
`

In terms of maintenance, this will not break anything that I know of - I have been using OSGi for a few years now, looking into the tooling and have not come across any reports of such. The contents of the added instructions would only need to change if new packages were referenced in calls to Class.forName() and in that case it would only be OSGi systems that would be broken because the metadata would not be up-to date; it would not affect non-OSGi users of your JAR file.

Your dependencies are actually already using OSGi - if you look into the Netty JAR files you will see that all of them apart from netty-all.jar have OSGi information in there already. Therefore you know that anybody using Pushy is already loading JAR files with OSGi information included. I hope that is some reassurance that this is a safe change!

jchambers:
Thank you for the thoughtful explanation! That's very helpful to me, and I'm fairly confident it makes sense to move forward with this change. Let me do a little more reading on my own, but I expect to merge this shortly.

arunan-interlink:
btw, another approach could be instead of storing class names, store the class object - e.g. Map<String, String>. The advantage would be then it is in the byte code, and so stuff like autowrapping would magically just work. It would change class resolution to the point that classes like ServerChannelClassUtil are loaded, but since you depend on Netty then that should not fail anyway. Doing that would remove the need for the Import-Package block in the pom. You would not in fact need to have any of this PR then, but OSGi fans such as I would prefer it :)

arunan-interlink:
I don't understand how the changes in this PR wouldhave caused that last build failure :/

arunan-interlink:
@jchambers I have rebased this PR onto current master and it builds fine. Are you OK to merge it please? It would mean I could use Pushy as-is rather than maintaining an OSGi-enabled variant.

jchambers:
I'm really sorry for the long delay here. If you can do a rebase onto master, I'll get this merged.

arunan-interlink:
Rebased, repushed.

I have noticed a "drop_oio" branch, which is something this branch refers to. Whichever of the branches gets merged first, the other would have to be updated.

I have also noticed a "java8" branch. This branch uses a version of maven-bundle-plugin that was chosen explicitly to support Java 7; this could be bumped if desired after that other branch was merged. Leaving it as it is would not be a problem, although removing the comment may be a good idea.

Thanks for getting back on this!

jchambers:
> I have noticed a "drop_oio" branch…
> I have also noticed a "java8" branch…

It turns out both were semi-inspired by this pull request. I'm definitely aware of the interactions and will take responsibility for ironing them out.

Thanks kindly for your work and patience!

PR #690, opened 2019-04-04
Implement factory methods to return HttpProxyHandlerFactory instances or not based on system default proxies

arunan-interlink:
The code has been structured to allow doing the same for SOCKS proxies in the same way, hence the new package-protected helper class.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/proxy/HttpProxyHandlerFactory.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/proxy/HttpProxyHandlerFactory.java
@@ -22,7 +22,12 @@
 
 package com.turo.pushy.apns.proxy;
 
+import java.net.Proxy;
 import java.net.SocketAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import com.turo.pushy.apns.ApnsClientBuilder;
 
 import io.netty.handler.proxy.HttpProxyHandler;
 import io.netty.handler.proxy.ProxyHandler;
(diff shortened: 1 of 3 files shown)


arunan-interlink:
This PR can be manually tested by setting the system properties
- https.proxyHost
- https.proxyPort
and for bonus points you can test varying by host with
- http.nonProxyHosts
(note http not https prefix)

This has been tested in Apache Karaf by modifying etc/system.properties to set them.

The code as written should be easy to copy and paste for SOCKS versions but I have nothing to run them against.

arunan-interlink:
The build failure is in different tests to the ones I added. I thought my tests had reset the global state they modify (the default proxy config) so I don't know what's happening there.

arunan-interlink:
Actually, a Proxy is just a wrapper around a SocketAddress and a Proxy.Type; the helper could just return the SocketAddress or null, saving the caller having to unwrap it. I'll fiddle with the patch ...

arunan-interlink:
Further commit added to have the proxy locator just return the appropriate SocketAddress. Either works, so which you would take (if any) is down to style choice really.

arunan-interlink:
My apologies, I believe I was premature here. I wrote this one blind as we are deploying to a customer behind an HTTP proxy. I watched their code startup and say that the factory had initialised as I had expected. However, it appears that Netty's proxy handler is HTTP/1 only so with Pushy now using HTTP/2 this does not work. We only got to test with the customer recently - I should have waited for a positive end to end test before submitting this PR.

jchambers:
> However, it appears that Netty's proxy handler is HTTP/1 only so with Pushy now using HTTP/2 this does not work.

So, let me begin by being clear that I am a total proxy n00b. When we first added proxy support, my thought was that a conventional "HTTP proxy" wouldn't work for this case precisely because APNs uses HTTP/2 (and prior iterations of APNs used a proprietary, non-HTTP protocol). Another user insisted that it would work because HTTP proxies can handle arbitrary traffic. Now that I'm looking at it in more detail, though, I can't find any evidence to support that claim.

If that's true (and I'll want to dig a little deeper before committing), it probably makes sense to remove the `HttpProxyHandlerFactory` entirely.

arunan-interlink:
I'm embarrassing myself in public as I also learn about HTTP proxies and HTTP/2!

My above comment was wrong; you can have an HTTP/1 proxy work with an HTTP/2 client as the proxy just creates a tunnel. HTTP/1 is used to talk to the proxy with a CONNECT message and then you send HTTP/2 which will be transparently forwarded down the tunnel. I have tested that here this morning using a Privoxy container.

The actual bug I was having was due to address returned from the locator not being resolved, which caused the proxy setup code to fail with what seemed to be a mysterious "unresolved host" exception. I have updated the branch with a fix for that.

Sorry for not writing this up already - I was waiting for a positive test at the customer site before I wrote anything else that I'd regret! I hope that should be done in a few hours. While I have proven the fix here, I don't yet know for sure that is the only problem the customer were having.

jchambers:
No worries at all; we're in the same boat ;)

I'll look forward to the test results!

PR #702, opened 2019-06-09
Add support for the `apns-push-type` header

### LINKED ISSUE #701: Support for apns-push-type
Via https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns?language=objc:

> `apns-push-type `
> (**Required when delivering notifications to devices running iOS 13 and later, or watchOS 6 and later.** Ignored on earlier system versions.) The type of the notification. The value of this header is `alert` or `background`. Specify `alert` when the delivery of your notification displays an alert, plays a sound, or badges your app's icon. Specify `background` for silent notifications that do not interact with the user.
>
> The value of this header must accurately reflect the contents of your notification's payload. If there is a mismatch, or if the header is missing on required systems, APNs may delay the delivery of the notification or drop it altogether.
>
> For information about configuring notification alerts, see [Generating a Remote Notification](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification?language=objc). For more information about sending background notifications, see [Pushing Background Updates to Your App](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/pushing_background_updates_to_your_app?language=objc).

Thanks,
Nick

jchambers:
This is a first pass at adding support for the `apns-push-type` header introduced in iOS 13/watchOS 6. There are a couple points of tension I'm trying to resolve (or accept):

1. `SimpleApnsPushNotification` now has a _lot_ of constructors. Doing everything in constructors made sense early on when there were only a couple optional fields, but the constructor count will grow exponentially as the APNs API adds more and more fields over time. In a future, larger release (where there's more expectation of API breakage), it might make sense to move to a builder pattern.
2. I'm not sure that I like making `pushType` be possibly-`null`. As a header, it falls into a grey zone of "required for successful operation sometimes, but ignored at other times." Making it nullable means minimal disruption and surprise for existing use cases. Making it required reduces surprise for new use cases.

I'm also wondering about the interplay of delivery priority and `apns-push-type`. Are all combinations of priorities and push types legal? Can you have a "conserve power" priority for an "alert" push type, for example?

I'm very open to feedback on all of this.

If merged, this closes #701.

----

TODO:

- [x] Un-hose existing tests.
- [x] Add more robust tests to make sure the `apns-push-type` header is actually being sent as expected.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
@@ -70,6 +70,7 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
     private static final AsciiString APNS_PRIORITY_HEADER = new AsciiString("apns-priority");
     private static final AsciiString APNS_COLLAPSE_ID_HEADER = new AsciiString("apns-collapse-id");
     private static final AsciiString APNS_ID_HEADER = new AsciiString("apns-id");
+    private static final AsciiString APNS_PUSH_TYPE_HEADER = new AsciiString("apns-push-type");
 
     private static final int INITIAL_PAYLOAD_BUFFER_CAPACITY = 4096;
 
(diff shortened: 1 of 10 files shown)


rbrightwell:
Looking forward to getting this fix.

PR #720, opened 2019-09-25
added VoIP PushType for VoIP push notifications

### LINKED ISSUE #719: iOS 13 apns-push-type should be "voip" for voip push
On iOS 13, VoIP pushes are getting rejected by APNs servers with the reason `InvalidPushType`.  AS mentioned [here](https://forums.developer.apple.com/thread/121083), we require to send `"apns-push-type" : "voip"` for VoIP pushes. But, it's not documented in the [official docs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns#2947610).

I've tested it locally by sending a push to my device on iOS 13 and it is indeed true. Would need to add that in `PushType` enum as without that, I'm unable to use this lib.

pankajsinghal:
On iOS 13, VoIP pushes are getting rejected by APNs servers with the reason `InvalidPushType`.  AS mentioned [here](https://forums.developer.apple.com/thread/121083), we require to send `"apns-push-type" : "voip"` for VoIP pushes. But, it's not documented in the [official docs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns#2947610).

I've tested it locally by sending a push to my device on iOS 13 and found the case to be true. Hence, added `voip` in `PushType` so that we can start using it for `VoIP` pushes for iOS>=13.

This closes #719.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/PushType.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/PushType.java
@@ -42,7 +42,12 @@ public enum PushType {
     /**
      * Indicates that a push notification is not expected to interact with the user on the receiving device.
      */
-    BACKGROUND("background");
+    BACKGROUND("background"),
+
+    /**
+     * Indicates that a push notification is expected to activate the client for handling VoIP flow.
+     */
+    VOIP("voip");
 
     private final String headerValue;
 


LowAmmo:
Would like to have this pulled in as soon as possible, as we have no idea what Apple's time line might be to start enforcing the 'apns-push-type' on the production APNS servers.

-Thanks!

jchambers:
@pankajsinghal Thank you for this contribution. Could you please add a description for posterity?

pankajsinghal:
@jchambers Have added the description. Should you require any changes, I'd be happy to help.

jchambers:
Folks, I've learned that there are actually several new values for the `apns-push-type` header, and expect the official docs to reflect all of the new values within the next couple days. Let's hold on this for the moment so we can get everything done in one shot.

Thanks for your patience!

pankajsinghal:
@jchambers 'd recommend we let it in and add new values when they are available. 

This is because APNS servers are not allowing VoIP pushes currently for iOS 13 without the above setting. And many of our users have already upgraded to the latest iOS version. Hence, we are currently facing issues in our production related to VoIP pushes.

LowAmmo:
@pankajsinghal - are you seeing issues when using the APNS Prod services?  We are actually only seeing problems with Debug builds of our application(s).  If we do a Distribution build and hit the Prod APNS services, everything seems to still be working.

LowAmmo:
As a note, on the Feedback Assistance Ticket that we logged to apple, we did get confirmation that 'voip' is the proper value for a VOIP push.

And they also mentioned (as @jchambers did), that updated documentation is coming.

jchambers:
[The official docs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns) have been updated. I'll add some commits tomorrow to cover all of the new values and link to the docs.

jchambers:
Tests were failing here due to some Travis CI changes that required a build matrix update on our end. I took care of that in #722.

@pankajsinghal We have two ways forward here:

1. You could rebase this onto `master` and I can merge it.
2. We can close this and I'll just cherry-pick your commit into the branch that covers all of the new `apns-push-type` values.

I'm fine either way; do you have a preference?

pankajsinghal:
@jchambers I've rebased anyway. Anything is fine with me. Either merge this PR or cherry-pick. Your call. 

Thanks for actively following up on this though. 👍

jchambers:
Thanks for rebasing. I've merged this and opened #723 to cover the remaining `apns-push-type` values.

Issue #725, opened 2019-10-14
The response of sendNotification may not fulfill


ylgrgyq:
Hi jchambers,

I thought I found another issue. Please take some time to review it when you are available.

It seems the response promise for [sendNotification](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java#L199) may stay in unfinished state forever when write failure occurs [here](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java#L228) or [here](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java#L224). Because there's no listener on `writePromise` to pass any exceptions it got to `responsePromise`. 

I think maybe we can fix this by changing [the block start at line 214 in ApnsClient to](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java#L214):
```java
channel.writeAndFlush(responsePromise).addListener(new GenericFutureListener<ChannelFuture>() {

    @Override
    public void operationComplete(final ChannelFuture future) throws Exception {
        if (future.isSuccess()) {
            ApnsClient.this.metricsListener.handleNotificationSent(ApnsClient.this, notificationId);
        } else {
            responsePromise.tryFailure(future.cause());
        }
    }
});
```
And removes the `tryFailure` [here](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java#L216).

What do you think? 

Sorry for not taking PR right now because I may be missing something. So if you agree with the modifications above, I would love to take a PR as soon as possible.

By the way, my enviroment is :
jdk: 1.8.0_192
pushy: 0.13.10
netty: 4.1.37.Final
tcnative: netty-tcnative-boringssl-static-2.0.25.Final-linux-x86_64

Thanks.

jchambers:
At a quick glance, I think you may be right. I'll analyze the situation more carefully as soon as I can. Thanks kindly for the report!

barkefors:
I would bumbly like to "bump" this, if time permits.

@ylgrgyq: Are you running with the PR changes in production and seeing a difference? We've recently started observing a similar behavior once our "batches" reached a certain volume.

ylgrgyq:
@barkefors : 
[...]

PR #727, opened 2019-10-14
Fulfuill response promise when write notification failed

ylgrgyq:
As for #725, I think I'd better write some codes to make myself more clear. Just words and links may sounds confused.


pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
```diff
@@ -342,6 +343,7 @@ private void handleEndOfStream(final ChannelHandlerContext context, final Http2S
                 this.handleErrorResponse(context, stream.id(), headers, pushNotification, errorResponse);
             } else {
                 log.warn("Gateway sent an end-of-stream HEADERS frame for an unsuccessful notification.");
+                // Shall we use a dedicated exception to fail responsePromise instead of just a log?
```
ylgrgyq: This is a question.

Issue #730, opened 2019-10-31
CVE-2019-16869 in netty dependency prior to 4.1.42.Final 


grossjonas:
This [CVE](https://nvd.nist.gov/vuln/detail/CVE-2019-16869) has a high score.
Current netty version is 4.1.43.Final

Please update.

jchambers:
The CVE doesn't seem relevant to Pushy; we're not running an HTTP server. We'll get this as part of a routine update, but I don't think there's a reason to do an emergency release.

jashaj:
Three other CVEs in Netty have been reported as well that come as a transitive dependency via pushy:

- https://nvd.nist.gov/vuln/detail/CVE-2019-20445
- https://nvd.nist.gov/vuln/detail/CVE-2019-20444
- https://nvd.nist.gov/vuln/detail/CVE-2020-7238

Any plans to update to the latest netty version and release pushy?

jchambers:
Again, none of these CVEs seem to directly affect Pushy. I would like to update and release, but will need to come back to this shortly as we're working out some project ownership questions right now.

jchambers:
We've updated to Netty 4.1.45 in #738.

PR #738, opened 2020-02-03
Update Netty to 4.1.45

DCKcode:
Hi,

This is a quick PR to update Netty to the latest upstream version. Netty is affected by [CVE-2019-20444](https://nvd.nist.gov/vuln/detail/CVE-2019-20444) and [CVE-2019-20445](https://nvd.nist.gov/vuln/detail/CVE-2019-20445). Although these are of course bigger concerns when Netty is used for server purposes, these CVEs also affect Netty when using it as an HTTP client.

I mostly want to reduce the noise users get from any vulnerability scanners they have running for future versions of Pushy :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ If you use [Maven](http://maven.apache.org/), you can add Pushy to your project 
 
 If you don't use Maven (or something else that understands Maven dependencies, like Gradle), you can [download Pushy as a `.jar` file](https://github.com/relayrides/pushy/releases/download/pushy-0.13.10/pushy-0.13.10.jar) and add it to your project directly. You'll also need to make sure you have Pushy's runtime dependencies on your classpath. They are:
 
-- [netty 4.1.37](http://netty.io/)
+- [netty 4.1.45](http://netty.io/)
 - [gson 2.6](https://github.com/google/gson)
 - [slf4j 1.7](http://www.slf4j.org/) (and possibly an SLF4J binding, as described in the [logging](#logging) section below)
 - [fast-uuid 0.1](https://github.com/jchambers/fast-uuid)
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you don't use Maven (or something else that understands Maven dependencies, l
 - [slf4j 1.7](http://www.slf4j.org/) (and possibly an SLF4J binding, as described in the [logging](#logging) section below)
 - [fast-uuid 0.1](https://github.com/jchambers/fast-uuid)
 
-Pushy itself requires Java 7 or newer to build and run. Under Java 7, Pushy has an additional dependency (included automatically by dependency management systems) on [netty-native 2.0.25.Final](http://netty.io/wiki/forked-tomcat-native.html), a native SSL provider that (among other benefits) includes ciphers required by APNs that are not included with Java 7 by default.
+Pushy itself requires Java 7 or newer to build and run. Under Java 7, Pushy has an additional dependency (included automatically by dependency management systems) on [netty-native 2.0.26.Final](http://netty.io/wiki/forked-tomcat-native.html), a native SSL provider that (among other benefits) includes ciphers required by APNs that are not included with Java 7 by default.
 
 Under Java 8 and newer, Pushy does not require a native SSL provider, but users may choose to use it regardless for enhanced performance. To use a native provider, make sure netty-tcnative is on your classpath. Maven users may add a dependency to their project as follows:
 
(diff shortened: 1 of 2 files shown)


jchambers:
Thanks for the contribution!

Could you please also update the README? I suspect there may be a matching `netty-tcnative` update, too.

As mentioned in #730, I don't think any of these CVEs actually affect Pushy. I also won't be able to merge this until some project ownership questions are resolved, but I expect that to get cleared up any day now.

DCKcode:
> Could you please also update the README?
Good catch, updated the version there too.

> I suspect there may be a matching netty-tcnative update, too.
Although that doesn't seem to have known security issues, I've pushed an update for that too.

> As mentioned in #730, I don't think any of these CVEs actually affect Pushy.
I agree the CVE in #730 doesn't affect Pushy, as it only affects server uses of Netty. The CVEs I mention in the PR actually do seem to affect Pushy, as they affect a class used for both server and client purposes. The risk is very small for Pushy's common use cases, but based on what I can see it is technically affected.

DCKcode:
Hmm - it seems the update of `netty-tcnative` is causing CI problems. Based on the logs I can't be sure if it's related to the patch, but I can't the check either.

Let me know if you need a revert.

jchambers:
> Although that doesn't seem to have known security issues

Right; acknowledged. Netty and netty-tcnative versions tend to be closely coupled, though, and there are often breaking changes on one end or the other. The band of netty-tcnative versions that works with a given version of Netty (and vice-versa) is historically very narrow and, in my opinion, not really well-documented.

> The CVEs I mention in the PR actually do seem to affect Pushy, as they affect a class used for both server and client purposes.

It's true that the affected class `HttpObjectDecoder` can be used in both client and server contexts, but Pushy doesn't use it at all. I'm not even sure if it's used by Netty's HTTP/2 stack at all (it may be, but not in any part used by Pushy!); HTTP and HTTP/2 are pretty different and share surprisingly few components within Netty.

jchambers:
> Based on the logs I can't be sure if it's related to the patch, but I can't restart the check either.

It seems fine to me in local testing, but I'm afraid I can't kick of a re-run either until those project ownership issues are ironed out. I'll press the issue internally.

DCKcode:
Ah - understood, thanks for clarifying. Hope you can iron out the issue!

jchambers:
Okay! I re-ran the failed tests aaaaaaand they're still failing. I'll have to do some digging to figure out which combination of Travis, Xcode, macOS, Netty, and netty-tcnative is causing the problem.

DCKcode:
Hmm, that might be getting hairy. I can revert the `tcnative` change if you prefer. I only added it because we had the impression it would be an easy thing to include, but it seems it might warrant its own PR anyway.

jchambers:
Well, let's not abandon it yet. We've found _something,_ and I'd like to understand what it is before we way forward. I'm trying to find a minimal crash reproducer in https://github.com/jchambers/netty-tcnative-travis-test and hope to have that settled shortly. With that under control, I think we can make an informed decision.

jchambers:
Well, I've narrowed it down substantially, still have plenty to do to get a truly minimal reproduction case. The problem started somewhere between netty-tcnative-2.0.26 and netty-tcnative-2.0.27. The version of Netty itself (as long as it's compatible with that version of netty-tcnative) doesn't matter, and so I'm reasonably convinced it's a tcnative issue. I know that was the suspicion from the beginning, but I'm more confident having eliminated some other possibilities.

The problem seems to happen somewhere when actually connecting to the server and sending a push notification. Just creating a server and a client doesn't cause the crash; they actually need to interact. That's also not entirely surprising, but again, it's good to eliminate possibilities.

I did most of this debugging in the [`netty-4_1_45-debug`](https://github.com/relayrides/pushy/tree/netty-4_1_45-debug) branch, and now will turn my effort back to https://github.com/jchambers/netty-tcnative-travis-test to try to build a more focused minimal reproduction case.

jchambers:
I've isolated the problem and opened an issue upstream: https://github.com/netty/netty-tcnative/issues/523

jchambers:
@DCKcode It looks like whatever went wrong happened in netty-tcnative 2.0.27. I think going for 2.0.26 in this pull request is the best way forward. Would you mind making the change? Thanks!

DCKcode:
Thanks for following through on this - I've put it back on 2.0.26.

jchambers:
Awesome. Thank you!

I hate to be picky, but there's a reference to the netty-tcnative version in the README, too; would you mind updating that, too? Sorry I didn't catch it earlier!

DCKcode:
Ah, of course. Updated.

jchambers:
Great—thanks very much!

For the record, version bumps aren't usually quite _this_ complicated.

PR #741, opened 2020-02-08
Prepare for change in project ownership

jchambers:
Friends, as those of you who subscribe to the [mailing list](https://groups.google.com/forum/#!forum/pushy-apns) may already know, we'll be transferring ownership of Pushy as a project from Turo as an organization to me directly in the near future. For more context and discussion, please see https://groups.google.com/d/msg/pushy-apns/U-CZdMiuMMI/0EeUhiWDBQAJ.

This set of changes does most—but not all—of the things we'll need to do to move Pushy to a new home soon. Notably, it does _not_ change the Maven coordinates for project artifacts in the various README files. To avoid confusion, we'll cover that once we've done a new release after the transfer.

I'll plan to merge this soon; please give a shout if you have any questions or concerns in the meantime!


filiphr:
I am reading this one just now.  Would be interesting to where the project can grow in the future.

Anyways, I just wanted to say thank you for taking up the project and continuing the work you do with Pushy

jchambers:
>  Would be interesting to where the project can grow in the future.

I'm happy to share some thoughts! Could you give me a little more direction, though? What's on your mind?

filiphr:
>Could you give me a little more direction, though? What's on your mind?

We are not that heavy users of the library, so I don't have many ideas. The current API for building the APNs client is really good and we were able to customize it to our needs quite easily. 

Here are some topics from me personally:

## Spring Boot

It might be interesting to know how the others from the community are using this. We are heavy Spring Boot users, so we needed to wrap it in an auto configuration in order to configure it via properties. From my point of view providing a Spring Boot auto configuration might be a good thing. However, this completely depends on the community, if there isn't big demand it doesn't make sense to go in that direction.

If there is interested into this I can see if we can do a contribution with the auto configuration.

## Integration

There are some issues from me personally. Those are mostly feedback and enhancement that could be done to make the integration a bit easier. However, I have to say again the API is quite permissive, so we could easily integrate what we needed.

## Reactive Stack

We are not really using it. However, I know that there is some push in that direction currently. Pushy is already more or less ready since it uses `Future`(s) (I think I am not that well versed in the internals)

jchambers:
@filiphr These all sound like important ideas, and I'd love to learn more about them. Would you mind raising these on the [mailing list](https://groups.google.com/d/forum/pushy-apns)? I'd hate for a robust discussion about future features to get lost because the pull request where they were raised got merged/closed.

Thanks!

Issue #466, opened 2017-05-12
Add per-notification timeouts


jchambers:
As a follow-up to #433, we should consider implementing per-notification timeouts.

dfroberts2:
+1

jchambers:
After lots of recent discussion, I'm pretty convinced this is something we should _not_ do. In basically all cases, it's a workaround for another bug elsewhere in Pushy. The right thing to do seems like it's pretty much always to just fix the bug rather than adding a workaround.

I'm going to close this for now, but welcome feedback and arguments that this provides non-workaround value.

Thanks!

PR #742, opened 2020-02-09
Attach write failure listeners unconditionally

### LINKED ISSUE #725: The response of sendNotification may not fulfill

jchambers:
This fixes #725 by attaching a listener to write futures unconditionally. Previously, we were only attaching the write failure if the channel were active and if streams were not exhausted, which could lead to abandoned write futures in some cases.

I'd _like_ to add tests to verify this behavior (both before and after the fix), but I'm reeeeeeeally struggling to appropriately mock things without basically rewriting the entire Netty HTTP/2 stack. I'll think on this a little more, but think it's safe to merge/ship this even without a test.

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
@@ -160,7 +160,20 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
     @Override
     public void write(final ChannelHandlerContext context, final Object message, final ChannelPromise writePromise) {
         if (message instanceof PushNotificationPromise) {
-            this.writePushNotification(context, (PushNotificationPromise) message, writePromise);
+            final PushNotificationPromise pushNotificationPromise = (PushNotificationPromise) message;
+
+            writePromise.addListener(new GenericFutureListener<ChannelPromise>() {
+
+                @Override
+                public void operationComplete(final ChannelPromise future) {
+                    if (!future.isSuccess()) {
+                        log.trace("Failed to write push notification.", future.cause());
+                        pushNotificationPromise.tryFailure(future.cause());
+                    }
+                }
+            });
+
+            this.writePushNotification(context, pushNotificationPromise, writePromise);
         } else {
             // This should never happen, but in case some foreign debris winds up in the pipeline, just pass it through.
             log.error("Unexpected object in pipeline: {}", message);
@@ -205,17 +218,6 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
                 final PromiseCombiner promiseCombiner = new PromiseCombiner();
                 promiseCombiner.addAll((ChannelFuture) headersPromise, dataPromise);
                 promiseCombiner.finish(writePromise);
-
-                writePromise.addListener(new GenericFutureListener<ChannelPromise>() {
-
-                    @Override
-                    public void operationComplete(final ChannelPromise future) {
-                        if (!future.isSuccess()) {
-                            log.trace("Failed to write push notification on stream {}.", streamId, future.cause());
-                            responsePromise.tryFailure(future.cause());
-                        }
-                    }
-                });
             } else {
                 // This is very unlikely, but in the event that we run out of stream IDs, we need to open a new
                 // connection. Just closing the context should be enough; automatic reconnection should take things


ylgrgyq:
Great wroks! Thanks  @jchambers.

And I found that there are some other cases which may lead to abandoned write futures, like when the programe goes to [here](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java#L304) or [here](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java#L344). So I think if I depend heavily on the futures returned from [sendNotification](https://github.com/relayrides/pushy/blob/master/pushy/src/main/java/com/turo/pushy/apns/ApnsClient.java#L199), I need a timeout on all these futures. If any of them not fullfill for too long, I should consider them failed and continue to do the following things. 

In our system, we have a rate limiting mechanism. Before sending a notification, we acquire a semaphore which will be released when the send future accomplished. Now I think I need the timeout facilities to release semaphore in case the future never complete.

jchambers:
> And I found that there are some other cases which may lead to abandoned write futures, like when the programe goes to here or here.

That's technically true, but those would both be cases where there's a serious upstream bug. Have you ever actually seen one of those things happen in production?

> So I think if I depend heavily on the futures returned from sendNotification, I need a timeout on all these futures.

Your point is taken (and this is certainly not the first time this request has come up—see #466), but I'm still skeptical that it's the right solution. There isn't really a mechanism to "cancel" a notification once it's been written, and so we could wind up with circumstances where a notification is marked as "failed" even though it was actually accepted by the server later. That said, we could explore making the `Future` returned by `sendNotification` cancellable up until it's actually written.

The real thing that makes me uneasy here is that we'd be adding a feature to Pushy to guard against bugs elsewhere in Pushy. It's an absurd analogy, if we had a `List` implementation and were afraid that `add` sometimes didn't work, it'd feel really weird to have a separate `add(T obj, int retryCount)` that just repeatedly tried to add the item to the list until it worked. The real thing to do would be to figure out why `add` wasn't working and just fix it rather than build features to work around it.

I think the same is true in this case; the right thing to do is just to find and fix the bugs rather than build a feature to work around them. Does that make sense to you?

Also, please accept my sincere apologies; I somehow completely missed that you had already opened #727 to address this issue! I do think I'd prefer to keep things in the handler instead of adding to `ApnsClient`, but that's feedback I would have been happy to provide on your pull request rather than just opening my own. My mistake!

ylgrgyq:
Thanks for your reply. 

> That's technically true, but those would both be cases where there's a serious upstream bug. Have you ever actually seen one of those things happen in production?

No, I have not seen it in production. It's just in my conjecture.

> I think the same is true in this case; the right thing to do is just to find and fix the bugs rather than build a feature to work around them. Does that make sense to you?

Yes. It's make sense on my second thought. Now I'm with you. 

> Also, please accept my sincere apologies; I somehow completely missed that you had already opened #727 to address this issue! I do think I'd prefer to keep things in the handler instead of adding to ApnsClient, but that's feedback I would have been happy to provide on your pull request rather than just opening my own. My mistake!

Please no worries. I provide #727 is only try to make myself more clear. It's not a kind of serious PR. I'm totally OK with #742 and I think this version is quite concise and elegant. Please feel free to merge it when you think it's all done.

Issue #696, opened 2019-05-01
Be able to configure Gson builder.


amir20001:
I'm personally running into an issue with date formatting but the same idea applied to any of the configs.

it would be nice if I could set my own Gson

jchambers:
This sounds like it's closely related to #678. We'll give it some thought.

amir20001:
@jchambers sorry about the delay. I would say the desired behaviour is pretty similar to #678, but I think the amount of work required to do it is quite different.

jchambers:
I've proposed #744 as a first step in a larger refactoring effort to make JSON serialization more flexible; feedback is welcome!

jchambers:
This is covered by #776.

Issue #678, opened 2019-01-29
Support for using Jackson instead of Gson


filiphr:
Currently the `ApnsPayloadBuilder` uses Gson to create the content for the payload. 

Is it possible to make the creation of payload pluggable via some kind of interface, or extract a base builder that would have common methods for the locations where the conversion to JSON is needed?

The reason why I am asking for this is due to the fact that I would like to try and use one JSON parsing library within our project and currently we use Jackson a lot.

I am also prepared to work on this if you are interested and would accept such a PR.

jchambers:
I think that's a reasonable suggestion, but I'd definitely want to find a way to do it that doesn't introduce more complexity when end users add Pushy as a dependency. A bad outcome, in my mind, would be having Pushy fail to produce payloads at runtime because somebody doesn't have any JSON libraries on their classpath.

Let me think on some strategies; I'm very open to suggestions!

filiphr:
In order to be backwards compatible I was thinking to continue having Gson as a compile dependency like you are doing now. And have an optional dependency on `Jackson` (unless you decide to force users to define a dependency). Otherwise, you have the risk that people will start to have problems, because you don't have a dependency. If you want you can provide a maven dependency that would exclude Gson and use Jackson in it (just a decorative pom).

In terms of implementation I was thinking to add something like `PayloadConverter` (naming can be whatever). 

That looks like:

```java
public interface PayloadConverter {

    // both Gson and Jackson support something like this
    String writeJsonPayload(Writer writer, Object value); 
}
```

Then there will be a `GsonPayloadConverter` and a `JacksonPayloadConverter`.

The `ApnsPayloadBuilder` will then have the payload converter field initialized in the constructor, and there can be an optional constructor where you can pass the converter. Additionally you can have some class that would provide the default converter based on the runtime classes. For example if you have Gson then Gson would be used, if you have Jackson then Jackson would be used. It is up to you to decide which one has preference (not sure which one is faster).

In the `ApnsPayloadBuilder` all the places that use the static `GSON` to just call the `payloadConverter#writeJsonPayload(Writer, Object)`.

jchambers:
Please excuse the long delay in getting back to this.

> I was thinking to continue having Gson as a compile dependency like you are doing now.

Respectfully, if we still have Gson as a dependency, why does it matter if we're using Gson or Jackson to generate the payload? Is the idea that you'd then add an `exclude` clause for Gson if you really, really didn't want it?

filiphr:
[...]

PR #744, opened 2020-02-14
Proposal: drop the notion of maximum payload lengths

jchambers:
So waaaaaaaay back at the beginning of time when we first introduced `ApnsPayloadBuilder`, the maximum length of an APNs payload was a mere 256 bytes. Payloads with fairly short messages and just a little bit of metadata could very, very easily overrun that limit, and so we designed `ApnsPayloadBuilder` with methods like `buildWithDefaultMaximumLength()` and `buildWithMaximumLength(int)`. We'd enforce those limits by shortening the contents of the `alert` text if possible.

These days, payloads can be up to 4,096 bytes (or 5,120 for VOIP notifications), and it's much less likely that the literal text of a notification is going to be the thing that's pushing notification size over the limit. With that in mind, I propose simplifying APNs payload construction by dropping the maximum length arguments in `ApnsPayloadBuilder`.

Pros:

- Greatly simplifies payload construction
- Has the potential to speed things up a bit
- Smooths the path to using alternative/customizable JSON serializers (see #678, #696)

Cons:

- Breaking API change (but as the README says, "the public API may change significantly before a 1.0 release")
- Users can construct larger-than-allowed payloads which will be rejected by the APNs server

Feedback is very welcome! I'd especially call on known JSON enthusiasts (@filiphr, @amir20001) for your input. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ final SimpleApnsPushNotification pushNotification;
     final ApnsPayloadBuilder payloadBuilder = new ApnsPayloadBuilder();
     payloadBuilder.setAlertBody("Example!");
 
-    final String payload = payloadBuilder.buildWithDefaultMaximumLength();
+    final String payload = payloadBuilder.build();
     final String token = TokenUtil.sanitizeTokenString("<efc7492 bdbd8209>");
 
     pushNotification = new SimpleApnsPushNotification(token, "com.example.myApp", payload);
--- a/benchmark/src/main/java/com/eatthepath/pushy/apns/ApnsClientBenchmark.java
+++ b/benchmark/src/main/java/com/eatthepath/pushy/apns/ApnsClientBenchmark.java
@@ -105,8 +105,8 @@ public class ApnsClientBenchmark {
             final ApnsPayloadBuilder payloadBuilder = new ApnsPayloadBuilder();
 
             for (int i = 0; i < this.notificationCount; i++) {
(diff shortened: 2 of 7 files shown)


filiphr:
This looks good to me @jchambers. 

I like the direction towards using alternative / customizable JSON serializers.

What I would suggest is to still check for the default size and throw an `IllegalArgumentException`. Perhaps add the ability to build with or without a check. This would still reduce the complexity within the builder. The benefit of doing this would be that the user won't be able to get rejected by the APNs server and would get feedback earlier (if they opt into it).

From the perspective of how we are using pushy it won't make any difference, and I would personally use the approach without a size check.

jchambers:
> What I would suggest is to still check for the default size and throw an `IllegalArgumentException`.

I hear you, but a few things make me hesitate:

1. This is an easy check for callers to do on their own if it's important to them to catch the issue before sending to the APNs server. By contrast, shrinking down the message content was a reasonably challenging problem and a service that (at the time!) many users probably needed.
2. Payload size limits have changed over time and also vary by context (VOIP and non-VOIP notifications have different payload size limits, for example). I'm concerned that hard-coding limits in the payload builder could put us in a position where the limits would quickly get outdated and report something as an error when it is, in fact, completely fine.

Still, it's all new territory to me—I'll keep thinking about it!

filiphr:
I had a better look at the current implementation. Yes you are right the check can really easily be done on the caller side. I had something else in mind (haven't been in this area for a while 😄).

I would say that given the current state of APNs (varying content size and large enough) it is OK to not do any checks, but rather build the payload as is and let the callers do the check.

PR #745, opened 2020-02-15
Drop support for OIO

jchambers:
I can't imagine that anybody was actually using OIO for anything. It's also been deprecated upstream, so now seems like as good a time as any to drop it.


arunan-interlink:
https://github.com/relayrides/pushy/pull/689 has a reference to OIO.

jchambers:
> #689 has a reference to OIO.

Yep—now that #689 has been merged, I've updated this pull request to match. Thanks!

jchambers:
I'm going to go ahead and merge this because it seems like it's really not controversial. If anybody objects, please leave a comment here and I'll happily revisit the issue!

PR #746, opened 2020-02-16
Proposal: drop support for Java 7

jchambers:
When we started the Pushy project seven years ago, we opted to target Java 7 as the minimum supported version of Java. Java 7 is no longer officially, publicly supported, and it'd be great to move on to at least Java 8 generally to stay up to date (though even Java 8 is pretty old at this point) and take advantage of new language features.

This pull request does the bare minimum we'd need to do to move to Java 8; if we adopt the move to Java 8, I'd expect to follow it up with pull requests to use JSR-310 date/time types, JUnit 5, `Optionals`, and lambdas where appropriate.

Still, I'd love to get feedback from the community. Do you still depend on Java 7? What would this change mean for you? If there's little to no demand for Java 7 support, we could make this change soon. Otherwise, I'll come up with a more detailed transition plan.

Thanks, everybody!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,6 @@ matrix:
         - os: linux
           jdk: openjdk8
 
-        - os: linux
-          dist: trusty
-          jdk: openjdk7
-          install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -pl '!micrometer-metrics-listener'
-          script: mvn test -B -pl '!micrometer-metrics-listener'
-
         - os: osx
           osx_image: xcode8.3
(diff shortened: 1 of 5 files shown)


arunan-interlink:
Everything we would run this on uses Java 8 or above.

jmason:
Likewise, we are fully Java 8 :)

jchambers:
…and here I was thinking this would be controversial ;)

marcelstoer:
> …and here I was thinking this would be controversial ;)

For something that has been EOL since April 2015? I'm not surprised 😉

PR #747, opened 2020-02-18
Proactively rotate authentication tokens before the server complains

### LINKED ISSUE #732: The delay of APNs's response is very high(even up to 700s)
Time: 12:00-17:00 UTC everyday
Location: China
The delay of APNs's response: The elapse from sending a notification to APNs Server to receiving a response from it.
Normal delay: 500ms
High delay: 10s-700s
Through put: 50/s

Description:
1. One of my servers' delay is very high (even up to 700 seconds), but another server's delay just slightly increase.
2. Even though the delay is very high, the responses indicate that the most of notifications was post successfully.
3. Restarting my push service will immediately reduce the delay.

jchambers:
This fixes #732 by proactively expiring authentication tokens before the server complains that the token has expired. We've learned that the server will throttle connections down to a single concurrent stream if the token expires, and the expectation is that clients should proactively rotate tokens.

@janzar @helle @extraSix does this look good to you folks?

### DIFF
--- a/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
@@ -85,7 +85,7 @@ class ApnsChannelFactory implements PooledObjectFactory<Channel>, Closeable {
         }
     }
 
-    ApnsChannelFactory(final SslContext sslContext, final ApnsSigningKey signingKey,
+    ApnsChannelFactory(final SslContext sslContext, final ApnsSigningKey signingKey, final long tokenExpirationMillis,
                        final ProxyHandlerFactory proxyHandlerFactory, final int connectTimeoutMillis,
                        final long idlePingIntervalMillis, final long gracefulShutdownTimeoutMillis,
                        final Http2FrameLogger frameLogger, final InetSocketAddress apnsServerAddress,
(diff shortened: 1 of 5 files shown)


janzar:
Maybe configurable time would be better with default 50? Overall looks good!

jchambers:
> Maybe configurable time would be better with default 50?

Good point. I'd gone back and forth on that. I hesitated because I didn't want to tempt people to make adjustments that could lead to weird behavior, but I think I can cover that in the docs. I'll update this pull request shortly.

jchambers:
> I'll update this pull request shortly.

Done. Feedback is welcome!

janzar:
> > I'll update this pull request shortly.
> 
> Done. Feedback is welcome!

Looks good, I'm waiting for release of v0.13.11, because for us it is easier to update, when artifact is in maven repo.

Thank you!

extraSix:
I encountered a problem related to the authentication token, but the performance is different from others. I am not sure if it is related to this bug.

Yesterday I updated 6 servers to use authentication token for one app's pushes. This morning, in one of the servers, all the pushes of the app could not get response from APNs, while the pushes of the other apps were normal. 
The 'Future' object will not be released untill a respond is received, so that the service is OOM. 
The problem was resolved after restarting the service.

I don‘t know whether @janzar @helle also encountered this problem.
To avoid memory leak caused by unresponsiveness, maybe you add a timeout mechanism to the ’Future' object: automatically release after timeout?

jchambers:
@extraSix Yes, that sounds like the same problem.

Even though it's been requested several times, I still don't think adding a timeout is the right solution.

It's always possible to exhaust memory by enqueuing notifications faster than Pushy can send them. This specific bug makes it easier to do so by greatly reducing the rate at which notifications are sent. Still, even in the best of circumstances, it's possible to "flood" memory with notifications. The recommended practice is to establish some kind of [flow control](https://github.com/relayrides/pushy/wiki/Best-practices#flow-control) to limit the number of notifications that are "in flight" at any given time.

PR #749, opened 2020-02-23
Untangle connection error handling

### LINKED ISSUE #693: Double promise mark warning in case of bad certificate
If caller provides "bad" certificate to Pushy (i.e. certificate that causes CERTIFICATE_AND_PRIVATE_KEY_MISMATCH or SSLV3_ALERT_CERTIFICATE_REVOKED SSL error) pushy logs following warning:

> WARN  [apns-push-6] [2019-04-11T12:07:55.911Z] [c.t.p.a.ApnsChannelFactory] [] [] [] [] Tried to mark promise as "failed," but it was already done.

It seems to be a design problem, but more important that service logs are filled by warnings that look to be pointless (this problem have to be resolved and/or logged by caller)

### LINKED ISSUE #698: UDP Ports leak on calling close on ApnsClient - 0.13.5
I've experienced this bug in prod after updating to new Pushy and could easily reproduce situation:

```
    @Test
    public void test1() throws Exception {
        final SimpleApnsPushNotification pushNotification
                = new SimpleApnsPushNotification("apns:xxxxxxxxxxxxxxxxxxxxxxxx",
                DEFAULT_TOPIC, "{}");

        NioEventLoopGroup group = new NioEventLoopGroup(1);
        while (true) {
            final ApnsClient client = new ApnsClientBuilder()
                    .setApnsServer(ApnsClientBuilder.DEVELOPMENT_APNS_HOST)
                    .setSigningKey(signingKey)
                    .setEventLoopGroup(group)
                    .build();

            client.sendNotification(pushNotification).await();
            client.close().await();

            Thread.sleep(1000);
        }
    }
```

this
$ while true; do netstat -lpn --udp|grep java; sleep 1; done
shows port leak on every iteration
Please see https://github.com/relayrides/pushy/issues/675 for more details or reopen it

jchambers:
Pushy's connection and handshaking process has changed a lot over the years as the protocol has evolved, as Pushy itself has become more advanced, and as I've learned new things. One consequence of all of that change is that establishing connections is generally getting more robust over time. Another is that we sometimes wind up with accumulated cruft that was mostly there to support old use cases and doesn't make sense any more.

This pull request eliminates some of the old cruft and, as a result, greatly simplifies the process of detecting, reporting, and handling errors that pop up when establishing new connections to an APNs server. The major changes:

1. We now build the entire pipeline up front rather than adding handlers in phases.
2. We've eliminated several "backstop" error-catching mechanisms.
3. As a consequence of the other changes, we're able to centralize exception handling in the `ApnsClientHandler`, which is the last handler in the pipeline.

@eugeneseppel I believe this fixes #693 by eliminating competition between error-reporting mechanisms.

This is a pretty significant change, so I'm going to save it for the next "major" release. Feedback is very, very welcome!

This whole thing is easier to read with [whitespace-only changes hidden](https://github.com/relayrides/pushy/pull/749/files?w=1).

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelFactory.java
@@ -75,17 +75,6 @@ class ApnsChannelFactory implements PooledObjectFactory<Channel>, Closeable {
 
     private static final Logger log = LoggerFactory.getLogger(ApnsChannelFactory.class);
 
-    @ChannelHandler.Sharable
-    private static class ConnectionNegotiationErrorHandler extends ChannelInboundHandlerAdapter {
-
-        static final ConnectionNegotiationErrorHandler INSTANCE = new ConnectionNegotiationErrorHandler();
-
-        @Override
-        public void exceptionCaught(final ChannelHandlerContext context, final Throwable cause) {
-            tryFailureAndLogRejectedCause(context.channel().attr(CHANNEL_READY_PROMISE_ATTRIBUTE_KEY).get(), cause);
-        }
-    }
-
     ApnsChannelFactory(final SslContext sslContext, final ApnsSigningKey signingKey, final long tokenExpirationMillis,
                        final ProxyHandlerFactory proxyHandlerFactory, final int connectTimeoutMillis,
                        final long idlePingIntervalMillis, final long gracefulShutdownTimeoutMillis,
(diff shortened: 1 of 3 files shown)


pushy/src/main/java/com/turo/pushy/apns/ApnsChannelFactory.java
```diff
@@ -116,58 +101,47 @@ public void exceptionCaught(final ChannelHandlerContext context, final Throwable
...
-                            }
-
-                            pipeline.addLast(new FlushConsolidationHandler(FlushConsolidationHandler.DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES, true));
-                            pipeline.addLast(new IdleStateHandler(idlePingIntervalMillis, 0, 0, TimeUnit.MILLISECONDS));
-                            pipeline.addLast(apnsClientHandler);
-                            pipeline.remove(ConnectionNegotiationErrorHandler.INSTANCE);
-
-                            channel.attr(CHANNEL_READY_PROMISE_ATTRIBUTE_KEY).get().trySuccess(channel);
-                        } else {
-                            tryFailureAndLogRejectedCause(channel.attr(CHANNEL_READY_PROMISE_ATTRIBUTE_KEY).get(), handshakeFuture.cause());
```
jchambers: This failure now makes its way to the `exceptionCaught` method in `ApnsClientHandler`.

PR #750, opened 2020-02-23
Drop `trace` calls that duplicate a frame logger's functionality

jchambers:
This is a minor bit of cleanup that removes some superfluous logging (even if it was at the `trace` level) that is essentially covered by a [`Http2FrameLogger`](https://pushy-apns.org/apidocs/0.13/com/turo/pushy/apns/ApnsClientBuilder.html#setFrameLogger(io.netty.handler.codec.http2.Http2FrameLogger)).

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
@@ -294,13 +294,11 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
 
     @Override
     public int onDataRead(final ChannelHandlerContext context, final int streamId, final ByteBuf data, final int padding, final boolean endOfStream) {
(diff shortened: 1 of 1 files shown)


pushy/src/main/java/com/turo/pushy/apns/ApnsClientHandler.java
```diff
@@ -390,7 +387,7 @@ public void onSettingsAckRead(final ChannelHandlerContext ctx) {
 
     @Override
     public void onSettingsRead(final ChannelHandlerContext context, final Http2Settings settings) {
-        log.trace("Received settings from APNs gateway: {}", settings);
+        log.debug("Received settings from APNs gateway: {}", settings);
```
jchambers: We're learning that settings changes are an important part of understanding some performance issues, and I think it makes sense to increase its visibility to `debug` (as a "minor lifecycle event").

PR #751, opened 2020-02-24
Also get credentials for HTTP proxies from system properties

jchambers:
As a follow-up to #691, this also gets usernames/passwords for HTTP proxies from system properties. @arunan-interlink your feedback is welcome and appreciated!


arunan-interlink:
It appears to make sense after a quick eyeball.

jchambers:
Cool. Thanks! We'll roll with it.

PR #756, opened 2020-03-05
Update to JUnit 5

jchambers:
As a follow-up to #746, this updates all of our tests to JUnit 5 and replaces [JUnitParams](https://github.com/Pragmatists/JUnitParams) with JUnit's own parameterized tests.


pushy/pom.xml
```diff
@@ -132,12 +132,13 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <argLine>-Dorg.slf4j.simpleLogger.defaultLogLevel=warn -Dio.netty.leakDetection.level=paranoid -ea</argLine>
-                    <properties>
+                    <!-- TODO Restore this in a JUnit5-compliant way -->
```
jchambers: I'm working on this, but the Surefire plugin isn't making it easy.

jchambers: Got it—turns out the answer was to stop trying to do this via Surefire and use a `TestExecutionListener` instead (see 5d71975e).

PR #757, opened 2020-03-06
Return CompletableFutures for public-facing operations

jchambers:
As a follow-up to #746, this pull request replaces all public-facing uses of [`io.netty.util.concurrent.Future`](https://netty.io/4.1/api/io/netty/util/concurrent/Future.html) with [`java.util.concurrent.CompletableFuture`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html). `CompletableFuture` is part of the stock Java SDK and offers all of the affordances—most importantly being able to add listeners—that Netty's `Future` offers.


jchambers:
I should also point out that this pull request highlights how much non-IO stuff we're inadvertently doing on IO threads. I'll open a follow-up pull request to add an "insulation" layer.

jchambers:
Also, I'm trying to figure out how to describe the action of calling `CompletableFuture#thenApply(...)`. "Add a listener" seems imprecise since `CompletionStage` is its own concept in the Java world, but "add a dependent completion stage" also feels a little obtuse for folks who aren't already familiar with `CompletableFuture`/`CompletionStage`. I'm rolling with "add a follow-up task" for now, but would love suggestions for alternative phrasing!

PR #775, opened 2020-05-02
Drop GSON as a dependency

jchambers:
Folks have expressed a preference to be able to use the JSON library of their choice when serializing JSON payloads. As a step down that road, this pull request drops our dependency on GSON entirely. It does so by introducing our own internal JSON parser (built with [JavaCC](https://javacc.github.io/javacc/)) and our own internal JSON serializer.

Some of you might think this is an excessive step and a textbook example of "not invented here" thinking, and you're probably right. On a personal note, it was interesting for me to learn more about building parsers, and it was a welcome distraction during a difficult time. On the merits, though, this really _does_ make it easier to avoid religious wars over JSON libraries by way of not really taking a side.

In terms of performance, benchmarks show our own JSON serializer to be somewhere between GSON and Jackson in terms of performance for Pushy's use cases. I still need to write some benchmarks for the parser.

It's important to note that this _does_ represent a change in behavior for the payload builder, and particularly in how we handle non-primitive objects added as custom properties. I'll be following this up with pull requests to make it easier to choose a specific JSON library for payload serialization, and those will surely be breaking API changes, so I hope that'll help folks notice that payload serialization has changed and may need attention.

Another important note: depending on how you're building Pushy, you may need to run `mvn generate-sources` to actually generate the parser from the JavaCC file.

This is a big change, and feedback is very welcome.


pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
```diff
@@ -332,7 +328,16 @@ private void handleEndOfStream(final ChannelHandlerContext context, final Http2S
...
             if (data != null) {
-                final ErrorResponse errorResponse = GSON.fromJson(data.toString(StandardCharsets.UTF_8), ErrorResponse.class);
+                ErrorResponse errorResponse;
+
+                try {
+                    errorResponse = ErrorResponse.fromMap(
+                            jsonDeserializer.parseJsonObject(data.toString(StandardCharsets.UTF_8)));
+                } catch (final ParseException e) {
+                    log.error("Failed to parse error reasponse: data.toString(StandardCharsets.UTF_8)");
+                    errorResponse = new ErrorResponse(null, null);
```
jchambers, line 338: Oops—meant to leave a TODO here to revisit this and make sure it's really the safest approach.

jchambers:
> I still need to write some benchmarks for the parser.

I wrote some benchmarks for the parser. The idea is to parse a bunch of randomly-generated error responses, because that's the only time clients actually parse JSON. Here's how things stacked up:

| Library | Throughput (operations/second) |
|---------|--------------------------------|
| GSON    |     1,179,635.567 ± 10,643.227 |
| Jackson |     2,382,039.407 ± 27,617.493 |
| Pushy   |       519,082.270 ± 64,789.183 |

So, yes, we're quite a bit slower than the competition when it comes to parsing. At the same time, I don't think that's actually a big deal in practical terms. Let's say Pushy is cranking through 50,000 notifications per second. Let's say 10% of those notifications will result in an error response from the server, which works out to 5,000 notifications per second.

GSON, which we were using before this change, could parse about a million error responses per second, which works out to about 1µs per error response. If we're half as fast, that means we're taking 2µs per error response. In other words, this change adds about 1µs per error response.

Putting it all together, that would mean that this change would add about 5,000µs (5ms) to the processing time for that batch of 50,000 notifications, which I think is pretty negligible. I am, of course, open to arguments to the contrary.

jchambers:
I did some tuning and have substantially improved the performance of our own parser in 
fcbdbac:

| Library | Throughput (operations/second) |
|---------|--------------------------------|
| GSON    |    1,079,586.362 ± 10,528.352  |
| Jackson |    2,255,141.586 ± 85,957.339  |
| Pushy   |      852,380.653 ±  7,198.073  |

It's still a little slower than GSON, but the gap is now much, much smaller.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
jchambers: Having fixed the glaring typos, I think this is fine. This is exactly what would have happened before, but now we have a logging statement to warn us if it ever happens.

PR #776, opened 2020-05-06
Add GSON/Jackson payload builders as sub-modules

### LINKED ISSUE #678: Support for using Jackson instead of Gson

### LINKED ISSUE #696: Be able to configure Gson builder.

jchambers:
This closes #678 and closes #696 by introducing GSON- and Jackson-backed payload builders in their own sub-modules. Pushy users will be able to add either (or neither) payload builder as a dependency as needed, and both allow callers to provide their own JSON serializers.

TODO:

- [x] Docs
- [x] READMEs
- [x] Examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Pushy's APNs clients maintain an internal pool of connections to the APNs server
 final SimpleApnsPushNotification pushNotification;
 
 {
-    final ApnsPayloadBuilder payloadBuilder = new ApnsPayloadBuilder();
+    final ApnsPayloadBuilder payloadBuilder = new SimpleApnsPayloadBuilder();
     payloadBuilder.setAlertBody("Example!");
 
     final String payload = payloadBuilder.build();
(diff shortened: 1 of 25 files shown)


jchambers:
JSON enthusiasts (@amir20001, @filiphr) your feedback is both welcome and appreciated.

filiphr:
Thanks for doing this @jchambers. I only managed to have a look at this today. It looks great. I also saw that you removed the gson direct dependency separately. This is great news, it makes it really powerful.

PR #783, opened 2020-06-04
Treat token expirations as fatal errors

### LINKED ISSUE #568: "TooManyProviderTokenUpdates"  received when sent a push notification to apns using .p8
I was using the pushy library.

"TooManyProviderTokenUpdates"  received when sent a push notification to apns using .p8 file.

[https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html#//apple_ref/doc/uid/TP40008194-CH11-SW1]
This document indicates that "TooManyProviderTokenUpdates" means The provider token is being updated too often.

But i am still confused,What does that mean ?
The provider token is being updated by what ?

What's right solution? Any reply is appreciated,Thanks very much.

jchambers:
As a spiritual follow-on to #747, this fixes #568 by simply closing connections when the server thinks tokens have expired.

My best idea right now is that the server simply doesn't let connections use new tokens after one has actually expired and continues to respond to all subsequent notifications with an "expired token" response. If that's the case, our best recourse is to just close the connection and open a new one.


jchambers:
To clarify: because we're proactively expiring tokens, we should never actually wind up in a situation where the token has expired. Closing the connection deals what should be an exceptional case rather than serving as the primary mechanism for rotating tokens.

PR #786, opened 2020-06-05
Update the Readme to reflect new getTokenInvalidationTimestamp usage

mohamedhafez:
Update the Readme to reflect that `getTokenInvalidationTimestamp()` now returns an `Optional`

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ try {
         System.out.println("Notification rejected by the APNs gateway: " +
                 pushNotificationResponse.getRejectionReason());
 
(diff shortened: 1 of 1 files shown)


mohamedhafez:
Just a little suggestion, it if things that change the public API like this could get a full version number bump, like to 0.15.0 in this case, that would make it easier on users of the library to know how careful they have to be when updating. just my two cents though and of course appreciate all the work either way!

jchambers:
Thank you for the contribution.

Yes, I usually try to avoid breaking changes within major releases. This was just a mistake.

PR #796, opened 2020-07-15
json parser fix

kissmikki:
fix of issue: https://github.com/jchambers/pushy/issues/795

### DIFF
--- a/pushy/src/main/java/com/eatthepath/json/JsonParser.java
+++ b/pushy/src/main/java/com/eatthepath/json/JsonParser.java
@@ -250,7 +250,7 @@ public class JsonParser {
                 if (stringBuilder.length() == 0) {
                     unescapedString = jsonString.substring(start, position);
                 } else {
-                    stringBuilder.append(jsonString, start, jsonString.length() - 1);
+                    stringBuilder.append(jsonString, start, position);
                     unescapedString = stringBuilder.toString();
                 }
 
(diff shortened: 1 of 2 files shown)


kissmikki:
> Also, may I ask how this came to your attention? While I agree that this is a bug, I think the only things Pushy should be parsing in production are APNs error responses. Did you encounter this issue while using Pushy in normal operation, or were you trying to do something fancier?

We have a set of notification artifacts that handle a various client notifications such as sending push to iOS. I used Pushy to create a new artifact that fits into our existing framework and since I saw it has a json parser I used it. I extract a key from the string encoded payload (which is generated by Pushy's PayloadBuilder) and noticesdthat the decoded values were different from the encoded ones.

Sure, I can move the test - when I'm back from holiday.

jchambers:
> Sure, I can move the test - when I'm back from holiday.

Don't sweat it. I'll just merge this and do a quick follow-up commit.

As soon as I can figure out why Travis isn't running tests, anyhow. Hm.

jchambers:
Well, the tests definitely passed: https://travis-ci.org/github/jchambers/pushy/builds/708254193. It looks like the results just didn't make it back to github. I'll go ahead and merge this. Thank you for the contribution!

jchambers:
> I'll just merge this and do a quick follow-up commit.

Done in 1639656a.

PR #813, opened 2020-09-26
Add more logging around creating and closing channels

jchambers:
We've had a couple issues where more logging around creating and closing channels could be helpful. This adds logging messages when channels are created or destroyed.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
@@ -152,10 +152,12 @@ class ApnsChannelPool {
                         ApnsChannelPool.this.allChannels.add(channel);
                         ApnsChannelPool.this.metricsListener.handleConnectionAdded();
 
+                        log.debug("Created channel {}", channel);
                         acquirePromise.trySuccess(channel);
                     } else {
                         ApnsChannelPool.this.metricsListener.handleConnectionCreationFailed();
 
+                        log.warn("Failed to create channel", future.cause());
                         acquirePromise.tryFailure(future.cause());
 
                         // If we failed to open a connection, this is the end of the line for this acquisition
@@ -220,11 +222,12 @@ class ApnsChannelPool {
         this.idleChannels.remove(channel);
         this.allChannels.remove(channel);
 
+        log.debug("Discarded channel {}", channel);
         this.metricsListener.handleConnectionRemoved();
 
         this.channelFactory.destroy(channel, this.executor.newPromise()).addListener(destroyFuture -> {
             if (!destroyFuture.isSuccess()) {
-                log.warn("Failed to destroy channel.", destroyFuture.cause());
+                log.warn("Failed to destroy channel {}", channel, destroyFuture.cause());
             }
         });
     }


pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
```diff
@@ -152,10 +152,12 @@ private void acquireWithinEventExecutor(final Promise<Channel> acquirePromise) {
                         ApnsChannelPool.this.allChannels.add(channel);
                         ApnsChannelPool.this.metricsListener.handleConnectionAdded();
 
+                        log.debug("Created channel {}", channel);
                         acquirePromise.trySuccess(channel);
                     } else {
                         ApnsChannelPool.this.metricsListener.handleConnectionCreationFailed();
 
+                        log.debug("Failed to create channel", future.cause());
```
dbyron0, line 160: Is this more important than debug?

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
```diff
@@ -152,10 +152,12 @@ private void acquireWithinEventExecutor(final Promise<Channel> acquirePromise) {
                         ApnsChannelPool.this.allChannels.add(channel);
                         ApnsChannelPool.this.metricsListener.handleConnectionAdded();
 
+                        log.debug("Created channel {}", channel);
```
dbyron0, line 155: Not sure how often this happens, but maybe it's infrequent enough for info?

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
```diff
@@ -220,6 +222,7 @@ private void discardChannel(final Channel channel) {
         this.idleChannels.remove(channel);
         this.allChannels.remove(channel);
 
+        log.debug("Discarded channel {}", channel);
```
dbyron0, line 225: Same question here.  Is this serious enough, or does it happen infrequently enough for info/warn/error?

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
jchambers: Thank you for the feedback. For context, please see the [logging guidelines in the README](https://github.com/jchambers/pushy#logging). I don't think this qualifies as an "important lifecycle event," and so am confident `debug` is the right level for this one.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
jchambers: Again, I think this is expected and doesn't rise to the level of an "important lifecycle event." Let's leave it at `debug`.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
jchambers: I think there's an argument to be made that this could be `warn` ("serious, but recoverable errors"). I think there's also an argument that it could be `debug` ("expected exceptions"). If this happens, callers will know because an attempt to send a push notification will fail with `future.cause()` as its cause; it's also reported in the metrics.

Let me think on it a bit.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
dbyron0: Sounds good.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
jchambers: Yeah, let's call this `warn`.

jchambers:
Travis reeeeeally seems to be struggling lately, but tests really are passing for this change and I'm going to go ahead and merge it.

PR #817, opened 2020-10-13
SNI Support

udayan9803:
This pull request adds SNI support for pushy-based applications running behind a transparent proxy utilizing SNI.  For example the Squid proxy peak and splice feature looks at the TLS Client Hello message and the SNI info found in the message.  It makes decisions based on the server name (endpoint whitelisting).  See also https://github.com/netty/netty/issues/9643.
If we don't add this fix, the Squid proxy will reject the request and when we make the api call as below, it just hangs forever on the call.
final PushNotificationResponse<SimpleApnsPushNotification> pushNotificationResponse = 
							sendNotificationFuture.get();


jchambers (approved):
This seems reasonable to me. I think there are a couple whitespace wrinkles to iron out, but I can handle that in a follow-up commit.

Thank you for the contribution!

jchambers:
Ugh—I'm afraid this is now out of date with master. I've resolved the conflicts, squashed the commits, and fixed the whitespace hiccups in #821. I'll close this pull request in favor of that one, but your name is still on the commit. Thank you for the contribution!

Issue #816, opened 2020-10-09
APNs server does not respond to some notifications


jchambers:
We've received reports that, starting on or around September 19, 2020, APNs servers have stopped responding to some notifications. From Pushy's perspective, this can look like a `Future` that never resolves (completion handlers are never called and calls to `.get()` time out or wait forever). Please see the [mailing list thread on this topic](https://groups.google.com/g/pushy-apns/c/adcJ6ZjupYw/m/HVxp8E-EAAAJ) for additional background and discussion.

From HTTP/2 frame logs, the problem appears to be that the server simply never sends a `HEADERS` (or `DATA`) frame in response to a push notification and never closes the HTTP/2 stream associated with the notification.

At this point, the goal is to identify some specific notifications affected by this problem. If you've encountered this issue, we're hoping to get the UUIDs (`apns-id`) and approximate timestamps of some affected notifications in the interest of sharing information upstream. Because the problem is that the server isn't responding, you'll need to assign your own `apns-id` values to outbound notifications (`SimpleApnsNotification` has a pair of constructors that accept an `apnsId` argument—using `UUID.randomUUID()` is recommended) to be able to uniquely identify which notifications are having this problem.

This issue is intended to consolidate a number of other reports on this topic, including #807, #814, and #815.

maelaouane:
Hi @jchambers , 

We are facing the same issue as stated above. Unfortunately, we do not set apnsId on our notification since we build them with `new SimpleApnsPushNotification(token, topic, payload, invalidationTime, priority, pushType, null, null)`. 

Nevertheless, we have message-id in the payload to follow a given message sent in APNS (I can DM you some examples on twitter if needed). 

We've got 3k+ notification timeout (5 seconds timeout on the get) on one cluster last Wednesday and we managed to workaround the issue by re-initializing the connection as follow: 
`client.close().await()`
and then recreating a new apns client with a builder as advised 
```
                clientBuilder = new ApnsClientBuilder()
                        .setApnsServer(host)
                        .setSigningKey(ApnsSigningKey.loadFromPkcs8File(file, team, key));
                client = clientBuilder.build();
```
Our environment: pushy: 0.13.10, jdk: 1.8

Issues started beginning of this week and concerns all iOS users (IOS 12, 13 and 14). 

Thanks!

jchambers:
> Nevertheless, we have message-id in the payload to follow a given message sent in APNS (I can DM you some examples on twitter if needed).

[...]

Issue #811, opened 2020-09-21
TooManyProviderTokenUpdates issue on v0.14.1


zsophy111:
This issue is related to this one: https://github.com/jchambers/pushy/issues/793

We are using Pushy for our APNS push notification in a complex enterprise application.
We originally used Pushy 0.13.10, but we continuously had native memory growth issue, so we upgraded the version to 0.14.1.
v0.14.1 solved the memory issue, however, we have met this TooManyProviderTokenUpdates issue, which we haven’t seen on v0.13.10.

We think this issue comes from this Pushy fix 747: https://github.com/jchambers/pushy/pull/747/commits/d42f6d77584de2b75f2821ff01f8f9d3a24dd1cd
In this fix, it proactively invalidates APNS auth token (set it to null) every 50 mins by default, which makes the APNS client to retrieve new tokens after invalidation. If one APNS client retrieve new auth token more than once in 20mins from APNS server, it will return this TooManyProviderTokenUpdates error.

In our application, we don’t create an APNS client per thread but our APNS client is shared among threads.
In ApnsClient doc: https://pushy-apns.org/apidocs/0.14/com/eatthepath/pushy/apns/ApnsClient.html
it says “APNs clients are intended to be long-lived, persistent resources. They are also inherently thread-safe and can be shared across many threads in a complex application. “ so I think we’re using it correctly.

So in this Pushy fix 747, perhaps when the timer fires and the token is set to null, more than one thread try and fetch a new token which pushes us over the limit. A fix could potentially be for the timer thread to fetch a new token and overwrite the old one rather than just setting it to null.

We can not rollback to v0.13.10 because of the memory leak issue, or older version because we need the iOS13 features. So we really rely on the new release to fix this TooManyProviderTokenUpdates error. Please let me know if you need more information for the fix. Looking forward to hearing back!

jchambers:
@zsophy111 If you turn on debug-level logging, `TokenAuthenticationApnsClientHandler` should log whenever it generates a new token: https://github.com/jchambers/pushy/blob/70fe90c1f8aba19f67d68ab5d8a99fbe56f80efe/pushy/src/main/java/com/eatthepath/pushy/apns/TokenAuthenticationApnsClientHandler.java#L108

If you haven't already, could you please turn on debug logging for (at least) `TokenAuthenticationApnsClientHandler` so we can learn more about how often the token is actually getting changed?

[...]

PR #820, opened 2020-10-21
Cancel ping futures when channels close

jchambers:
Tangentially-related to #811 and #816, this fixes an issue where ping futures could stay in play (and possibly keep a handler alive) longer than expected/desired. I think this was leading to some strange log output from threads that were no longer actually serving a channel.


jchambers:
Travis really doesn't seem interested in running builds today, but tests pass and I'm going to go ahead and merge these changes.

PR #822, opened 2020-10-21
Log which channels are affected by token changes

jchambers:
To support the debugging effort for #811, this change adds the name/ID of affected channels in token change messages.


jchambers:
Travis is stiiiiiiiill struggling, but tests pass locally and so I'll go ahead and merge these changes.

PR #825, opened 2020-10-27
Initial GitHub Action workflow for testing pushy against various JDK versions

geertjanw:


jchambers:
Thank you; I missed this when you opened it months ago, and appreciate the contribution!

jchambers:
I've cherry-picked this into #848. Getting the build to a more solid place took a few extra steps (some of the build profiles were just flat-out wrong), and we need to drop Travis entirely, too, so I figured it made sense to consolidate everything into one pull request so it's clearer for posterity.

Thanks again!

PR #836, opened 2020-12-14
Catch GOAWAY message

arjenvdhave:
Catch GOAWAY message and fail the pending promise so the next push message will try and create a new channel

Fix for:
https://github.com/jchambers/pushy/issues/831


jchambers:
Thank you very much for the contribution, but I think we have this covered in #854.

PR #850, opened 2021-02-21
Share authentication tokens between connections within a client

### LINKED ISSUE #811: TooManyProviderTokenUpdates issue on v0.14.1

jchambers:
In an effort to fix #811, this introduces an `AuthenticationTokenProvider` that allows multiple connections within an `ApnsClient` instance to share a single authentication token, reducing the number of token rotations over time.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelFactory.java
@@ -22,7 +22,7 @@
 
 package com.eatthepath.pushy.apns;
 
-import com.eatthepath.pushy.apns.auth.ApnsSigningKey;
+import com.eatthepath.pushy.apns.auth.AuthenticationTokenProvider;
 import com.eatthepath.pushy.apns.proxy.ProxyHandlerFactory;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.*;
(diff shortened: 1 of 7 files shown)


jchambers:
One problem I spotted in review: `TokenAuthenticationApnsClientHandler` needs a way to report to the provider that it thinks the token has expired if we get a `TooManyProviderTokenUpdates` response from the server.

PR #851, opened 2021-02-22
Add methods with fallback if loc-key not found to ApnsPayloadBuilder

Anddor:
While for most cases only setting one or the other of loc-key and body (and likewise for title, subtitle), we wish to be able to set body and title explicitly along with loc-keys to be backwards-compatible with users who have not yet updated their apps with new localization file. Does this seem like a sensible solution to the issue?

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
@@ -179,6 +179,31 @@ public abstract class ApnsPayloadBuilder {
     }
 
     /**
+     * <p>Sets the key of a message in the receiving app's localized string list to be shown for the push notification,
+     * along with a fallback for apps that do not have the given string in the string list.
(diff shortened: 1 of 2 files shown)


dtechen:
Changes look good to me. So I'd really appreciate if they could go in.

Anddor:
Hi @jchambers, I think we are many users of this library who need fallback-behaviour to be backwards compatible. Does this solution look ok? Alternatively, removing the explicit null reset when setting both with the builder?

jchambers:
Hello, and thank you for the contribution. I'll review this as soon as I'm able, but I'm focused on some issues that I understand to be a higher priority right now.

KurtStauffer:
I'm also interested in getting this change out. Thanks!

jchambers:
Thank you for the contribution, but I think I'd rather just change the behavior of the existing methods to not be mutually-exclusive. I recognize that's a breaking change, but we have a couple of those on deck for the next release anyhow and I think keeping things simple is the right move here.

I'll close this for now and follow up with another pull request momentarily.

jchambers:
Please see #897 for a proposed implementation that allows for both localized and literal content.

PR #854, opened 2021-03-03
Fix #853: Reject inactive pending create channel

### LINKED ISSUE #853: Creating Channel stucks on the unstable network
Version: 0.14.2
Severity: Critical
Symptom: `ApnsChannelPool` cannot acquire `Channel` forever
Properties: concurrentConnections=1, eventLoopGroup=NioEventLoopGroup(1 thread)
Region: China(once every 2 days), Europe(once every 7 days)

Hi, Jon. I'm considering applying Pushy in the global environment. And I really appreciate your efforts.
Sadly I often faced a problem above with this so I'm trying to find the cause.

I believe these are the correct steps to reproduce in debug mode:
1. Break at the last line of `io.netty.channel.nio.AbstractNioChannel#connect()`
2. Down the network (ifconfig down or whatever)
3. Resume application
4. After few(<30?) seconds later, Netty calls `ApnsClientHandler#channelInactive()`
5. Resume the network
6. Then `ApnsChannelPool#pendingCreateChannelFutures` have 1 `createChannelFuture` that has not been done
7. `ApnsChannelPool#acquireWithinEventExecutor()` can no longer create the proper Channel

Of course, we can reproduce this symptom in normal mode.

I think we must finish the `createChannelFuture` so I suggest inserting lines on `ApnsClientHandler#channelInactive()` to resolve pending `createChannelFuture`.

I look forward to hearing from you. Thanks.

KimTaehee:


pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
```diff
@@ -507,6 +507,11 @@ public void channelInactive(final ChannelHandlerContext context) throws Exceptio
         }
 
         this.unattachedResponsePromisesByStreamId.clear();
+
+        if (!getChannelReadyPromise(context.channel()).isDone()) {
+            log.debug("Failed to create channel due to inactivity");
+            getChannelReadyPromise(context.channel()).tryFailure(STREAM_CLOSED_BEFORE_REPLY_EXCEPTION);
+        }
```
jchambers, line 514: Minor: we could simplify this to:

```suggestion
        if (!getChannelReadyPromise(context.channel()).tryFailure(STREAM_CLOSED_BEFORE_REPLY_EXCEPTION)) {
            log.debug("Failed to create channel due to inactivity");
        }
```

jchambers (commented):
Hello, and thank you for the contribution (and for the bug report in #853)!

I need a little more time to understand the root cause of this issue in detail, and will come back to this shortly.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
KimTaehee: It's nice. I applied. :+1:

jchambers:
Thank you for your patience! I agree that this is the right fix, and appreciate the contribution!

PR #886, opened 2021-05-25
Possibility to use externally managed SSL context. 

garikjan:
Please consider adding possibility to use externally managed SSL context with ApnsClientBuilder. E.g. to be able to implement custom server certificate validation policy, using custom TrustManagerFactory.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java
@@ -33,6 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.net.ssl.SSLException;
+import javax.net.ssl.TrustManagerFactory;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -67,6 +68,8 @@ public class ApnsClientBuilder {
     private File trustedServerCertificatePemFile;
     private InputStream trustedServerCertificateInputStream;
     private X509Certificate[] trustedServerCertificates;
+
+    private TrustManagerFactory trustManagerFactory;
 
     private EventLoopGroup eventLoopGroup;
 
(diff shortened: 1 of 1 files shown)


jchambers:
Can you please say more about your use case? What are you trying to do that's not covered by [`ApnsClientBuilder#setTrustedServerCertificateChain`](https://pushy-apns.org/apidocs/0.14/com/eatthepath/pushy/apns/ApnsClientBuilder.html#setTrustedServerCertificateChain(java.io.File))?

garikjan:
We need to intercept certificate changes - when new certificate is issued by Apple, so that we could consider that certificate as trusted in some cases to avoid downtime in our services. Recently Apple changed APNS certificate and we had several hours downtime in our push notification service, until we added new certificate to truststore.

jchambers:
> Recently Apple changed APNS certificate…

To be clear, do you mean the [March 29 transition away from GeoTrust certificates](https://developer.apple.com/news/?id=7gx0a2lp)?

garikjan:
Yes, unfortunately we missed that notification and didn't add new certificate to truststore... :( And we faced connection problem later, on 2nd of April, when probably reconnection was issued. So, we would like to avoid such problems in future, that's why we are looking for a solution.

garikjan:
Meanwhile I found simpler solution - just to add a setter:
setTrustManagerFactory(TrustManagerFactory trustManagerFactory) to ApnsClientBuilder class. And use in with SslContextBuilder:

                final SslContextBuilder sslContextBuilder = SslContextBuilder.forClient()
                        .trustManaget(trustManagerFactory)
                        .sslProvider(sslProvider)
                        .ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE);

jchambers:
Understood; thanks for the additional background!

While I certainly appreciate what you're trying to do, I think this approach carries risks that outweigh the benefits. In particular, I think this opens the door to folks following the widely-issued (and terrible) advice to just provide a "trust everything" `TrustManager` if they have certificate difficulties.

I think we have lots of other ways to address the problem going forward (like specifying multiple trusted certificates), and would prefer to pursue one of those instead.

Still, thank you very much for the contribution! It's greatly appreciated.

garikjan:
It was the requirement of our customer to completely eliminate such problems with certificates in the future. We will try to convince the customer that such a solution to the problem can lead to the worse consequences. Thank you very much for your participation and useful recommendations!

PR #894, opened 2021-06-24
Added support for iOS 15 interruption level

s1g53gv:
Add support for iOS 15 interruption-level attribute. Solves https://github.com/jchambers/pushy/issues/893.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
@@ -72,6 +72,8 @@ public abstract class ApnsPayloadBuilder {
     private String summaryArgument = null;
     private Integer summaryArgumentCount = null;
 
+    private String interruptionLevel = null;
+
     private String[] urlArguments = null;
 
     private final HashMap<String, Object> customProperties = new HashMap<>();
@@ -90,6 +92,7 @@ public abstract class ApnsPayloadBuilder {
     private static final String SUMMARY_ARGUMENT_KEY = "summary-arg";
     private static final String SUMMARY_ARGUMENT_COUNT_KEY = "summary-arg-count";
     private static final String URL_ARGS_KEY = "url-args";
+    private static final String INTERRUPTION_LEVEL_KEY = "interruption-level";
 
     private static final String ALERT_TITLE_KEY = "title";
     private static final String ALERT_TITLE_LOC_KEY = "title-loc-key";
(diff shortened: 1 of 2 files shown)


pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -115,6 +118,26 @@
      */
     public static final String DEFAULT_SOUND_FILENAME = "default";
 
+    /**
+     * The passive interruption level.
+     */
+    public static final String INTERRUPTION_LEVEL_PASSIVE = "passive";
```
jchambers, line 124: I see where you're going, but I think it might make sense to turn this into `enum InterruptionLevel` instead.

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -605,6 +628,26 @@ public ApnsPayloadBuilder setSummaryArgumentCount(final Integer summaryArgumentC
...
+     *
+     * <p>Accepted values are {@link INTERRUPTION_LEVEL_PASSIVE}, {@link INTERRUPTION_LEVEL_ACTIVE}, {@link INTERRUPTION_LEVEL_TIME_SENSITIVE} and {@link INTERRUPTION_LEVEL_CRITICAL}.</p>
+     *
+     * @param interruptionLevel the interruption level.
+     *
+     * @return a reference to this payload builder.
+     *
+     * @see @see <a href="https://developer.apple.com/design/human-interface-guidelines/ios/system-capabilities/notifications/">Notifications</a>
+     *
+     * @since 0.14.3
```
jchambers, line 642: I think the next release is actually going to be 0.15; can we change this to match?

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -115,6 +118,26 @@
      */
     public static final String DEFAULT_SOUND_FILENAME = "default";
 
+    /**
+     * The passive interruption level.
```
jchambers, line 122: For all of these levels, can we provide more specific docs? I think it's okay to quote the main docs as long as we provide attribution. For example:

```
According to <a href="...">Apple's documentation</a>, notifications with a {@code passive} interruption level contain "information people can view at their leisure, like a restaurant recommendation."
```

jchambers (changes requested):
Thank you for the contribution! I think this needs a couple small adjustments, but we should be able to get this merged and out the door shortly.

Cheers!

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
s1g53gv: Of course!

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
s1g53gv: The enum `InterruptionLevel` now reports Apple's HIG description for each level.

s1g53gv:
@jchambers I'm unsure if you have notifications enabled for the comments, so I'll bump you here too: let us know if the change meet your expectations :)

jchambers (approved):
I have a few small quibbles, but I think the core of this makes sense and the easiest way to resolve the quibbles is to just do a quick follow-up commit. Thanks for the contribution!

PR #930, opened 2021-12-11
pom: clarify description

simon04:


jchambers:
Seems reasonable. Thanks for the contribution!

PR #946, opened 2022-02-12
Return to a one-token-per-channel model

### LINKED ISSUE #931: TooManyProviderTokenUpdates issue on v0.15.1
Currently we use  pushy v0.15.1 for our APNS push notification in a enterprise application and receive "TooManyProviderTokenUpdates" sporadically when sent a push notification to apns using .p8 file. 

Below is the code to instance the ApnsClient

```
	private synchronized void activateApnsClient(final ConfigData activeConfigData) throws SSLException {
		if (activeConfigData != loadedConfigData) {
			closeApnsClient();
			loadedConfigData = activeConfigData;
			apnsClient = new ApnsClientBuilder()
					.setApnsServer(activeConfigData.getApnsHostName(), activeConfigData.getApnsServerport())
					.setSigningKey(activeConfigData.getApnsSigningKey())
					.setConnectionTimeout(Duration.ofSeconds(activeConfigData.getApnsConnectionTimeout(), 0))
					.setTrustedServerCertificateChain(activeConfigData.getTrustedServerCertFile())
					.build();
		}
	}

```
where ApnsClientBuilder.PRODUCTION_APNS_HOST and 2197 are used as Hostname and port for ApnsServer. Connection timeout is set to 5sec.

The ApnsClient will be closed and reinstantiated, when the configuration data are changed, what but seldom happens.

In our application the above code and pushy  v0.14.2 are used untill 23. nov. 2021without any problem. On 24. nov. 2021 we upgraded pushy to v0.15.1using netty 4.1.69 (see #922 ).  Since that time we get the response with "TooManyProviderTokenUpdates"  sometimes.

Any reply is appreciated. Thanks very much.

jchambers:
This essentially reverts #850 (though a direct revert wasn't practical given the volume of changes between #850 and the present). We've observed in #931 that a shared token model can lead to a situation where:

1. Channel A is opened and prompts generation of a new authentication token.
2. Time passes.
3. Channel B is opened shortly before the original authentication token is expired.
4. Channel B sends a message with the original authentication token.
5. Time passes.
6. The original authentication token expires and a new one is generated.
7. Channel B sends a message with the new authentication token.

In that case, from the server's perspective, channel B has used two different authentication tokens in rapid succession, leading to a `TooManyProviderTokenUpdates` rejection.

The hope is that this fixes #931. At the same time, this will need some testing because this is reverting to exactly(?) the same state that prompted us to try #850 in the first place.


kicktipp:
It still works fine after over a week of sending lots and lots of messages. Can you make a release?

jchambers:
Thank you for the update. I plan to issue a new release soon, but it's a process more limited by Real Life Stuff than any technical concern at this point.

PR #948, opened 2022-02-27
Handle multiple sets of credentials with a single ApnsClient instance

### LINKED ISSUE #540: Handle multiple sets of credentials with a single ApnsClient instance

jchambers:
Implementing multiple sets of credentials within a single client (which will close #540) will be no small feat. For now, this is an integration branch that serves two purposes:

1. It provides a home for a big, wide-ranging development project without making undue noise in `main`
2. It provides visibility into the effort

In time, this will become a more serious "for actual review" branch, but for now, it's just a draft and a bit of a holding area.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelFactory.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelFactory.java
@@ -116,12 +116,10 @@ class ApnsChannelFactory implements PooledObjectFactory<Channel>, Closeable {
                         clientHandlerBuilder = new TokenAuthenticationApnsClientHandler.TokenAuthenticationApnsClientHandlerBuilder()
                                 .signingKey(clientConfiguration.getSigningKey().get())
                                 .tokenExpiration(clientConfiguration.getTokenExpiration())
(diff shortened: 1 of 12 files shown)


jchambers:
I'm not exactly sure how I managed to close this branch, but something clearly went wrong while rebasing. Working on it.

jchambers:
Having made more progress on this implementation, I'm beginning to think it may not be the best idea. Please see https://github.com/jchambers/pushy/issues/540#issuecomment-1873033128 for discussion; feedback is very welcome!

jchambers:
As discussed in #540, let's pursue other paths to making Pushy more effective for multi-credential users.

PR #967, opened 2022-06-15
ApnsPayloadBuilder check custom property key not `aps`

liuchong:
A check to avoid accidentally overwriting the important `aps` field.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
@@ -704,8 +704,13 @@ public abstract class ApnsPayloadBuilder {
      * @param value the value of the custom property
      *
      * @return a reference to this payload builder
+     *
+     * @throws IllegalArgumentException if the key equals to "aps"
      */
     public ApnsPayloadBuilder addCustomProperty(final String key, final Object value) {
+        if (APS_KEY.equals(key)) {
+            throw new IllegalArgumentException("Custom property key must not be aps");
+        }
         this.customProperties.put(key, value);
         return this;
     }


pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -704,8 +704,15 @@ public ApnsPayloadBuilder setUrlArguments(final String... arguments) {
      * @param value the value of the custom property
      *
      * @return a reference to this payload builder
+     *
+     * @throws IllegalArgumentException if the key equals to "aps"
+     *
+     * @since 0.15.2
```
jchambers, line 710: I'm torn on this; this is technically a breaking change, but this method has been around much longer. I'm inclined to leave out the `@since` tag in this case, but am open to other ideas!

jchambers (approved):
Thanks for the contribution!

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
liuchong: Ok, I've removed the `@since` tag 👀

jchambers (approved):

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
jchambers: I'll make sure to call it out in the release notes, but agree that isn't entirely ideal. We're getting close to 1.0, though, and I'm planning on getting much stricter about semantic versioning when we get there.

liuchong: Thanks, it'll be great! 😍

PR #969, opened 2022-06-26
Update `jackson-databind` to 2.13.3

jchambers:
This updates the Jackson payload builder's `jackson-databind` dependency to 2.13.3 to address CVE-2020-36518. CVE-2020-36518 is a vulnerability that can be exploited by an attacker by sending a malicious payload to a Jackson JSON _parser,_ but Pushy only ever uses Jackson to _write_ JSON, and so we're not actually vulnerable. Still, seems wise to update.


jchambers:
Also worth noting: in cases where we _are_ parsing JSON, we're not generally processing input from untrusted users, but from Apple's servers.

PR #990, opened 2022-09-28
Support Live Activity push notifications

### LINKED ISSUE #975: iOS Live Activities
Apple just released Live Activities: https://developer.apple.com/documentation/activitykit/displaying-live-data-on-the-lock-screen-with-live-activities

This will allow servers to send push notifications directly to a live activity (as opposed to the app itself): https://developer.apple.com/documentation/activitykit/activity/pushtoken

I'm still catching up on the documentation, not sure if Pushy will need to support this in any way. It's possible that just using those tokens Pushy will work for this out of the box though.

ifrins:
Add support for Live Activity push notifications based on the [existing public docs](https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications)

The JavaDoc comments are still a bit TBD since the docs on Apple end are very light on details. 

This implementation has been tested on iOS 16.1 beta 3

This closes #975.


jchambers:
Friends, I know this is important to many of you. I'll review it as soon as I can, but I can't make it my top priority for at least a few more days.

NachoSoto:
I've been using this the last few days and it works great :)

ifrins:
I have been using this PR for sending Live Activity related pushes for the last few weeks and everything seems to work (including in the 16.1 RC).

ifrins:
Actually, there’s an issue with this PR and setting a sound for Live Activities. On _normal_ push notifications the sound is in `$.aps.sound` (and silent if it is not present). However for Live Activities, sound is in `$.aps.alert.sound`. 

I think I will create a separate Payload Builder for Live Activities otherwise it will be confusing to have different methods to set the sound and only one working.

Morbix:
Hi @jchambers, seems that the Live Activities feature will be released next Monday. 🤩 

Is there any chance to have this PR released in the next few days?

NachoSoto:
Not sure about the new approach. `LiveActivityApnsPayloadBuilder` is a brand new builder, which means we can't compose it with others like `GsonApnsPayloadBuilder`. We would need `GsonLiveActivityApnsPayloadBuilder` now.

NachoSoto:
Never mind that was easy to do:
```java
public final class GsonLiveActivityApnsPayloadBuilder extends LiveActivityApnsPayloadBuilder {
    private final Gson gson;
    private static final Gson DEFAULT_GSON = (new GsonBuilder()).disableHtmlEscaping().serializeNulls().create();

    public GsonLiveActivityApnsPayloadBuilder() {
        this(DEFAULT_GSON);
    }

    public GsonLiveActivityApnsPayloadBuilder(Gson gson) {
        this.gson = (Gson) Objects.requireNonNull(gson);
    }

    public String build() {
        return this.gson.toJson(this.buildPayloadMap());
    }
}
```

ifrins:
Makes sense @NachoSoto! I have actually added implementations for Jackson and Gson in this last commit. I'm using the default serialization so I hadn't noticed 👍

NachoSoto:
@jchambers I know life can be busy (I have several open source projects myself that I honestly don't have time to work on), I was just wondering if you had a chance to look into this. Thanks!

jchambers:
Not yet, but soon, I promise. I emphasize that I understand that this is important to many of you, and I will get to it as soon as I can.

pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityApnsPayloadBuilder.java
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Sets the literal text of the alert message. The expanded view of the Live Activity will be shown for a few
+     * seconds if an alert body and alert title are sent. It is unclear if the actual text is used anywhere.
+     *
+     * @param alertBody the literal message to be shown for this push notification
+     *
+     * @return a reference to this payload builder
+     */
+    public LiveActivityApnsPayloadBuilder setAlertBody(final String alertBody) {
```
NachoSoto, line 77: This will need to support localization too: https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications

ifrins: You're right, fixed!

jchambers:
> Actually, there’s an issue with this PR and setting a sound for Live Activities. On normal push notifications the sound is `in $.aps.sound` (and silent if it is not present). However for Live Activities, sound is in `$.aps.alert.sound`.

Reading the [Live Activity docs](https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications), I think I understand where this suggestion is coming from. At the time of writing, the docs include this example payload:

```json
{
    "aps": {
        "timestamp": 1168364460,
        "event": "update",
        "content-state": {
            "driverName": "Anne Johnson",
            "estimatedDeliveryTime": 1659416400
        },
        "alert": {
            "title": "Delivery Update",
            "body": "Your pizza order will arrive soon.",
            "sound": "example.aiff" 
        }
    }
}
```

I think there are two ways to interpret this:

1. Apple has introduced a new and otherwise undocumented requirement for the topology of sounds in the existing alert system that only applies when sending Live Activity updates or
2. Somebody made a typo in the example.

In the absence of other information, I'm inclined toward 2 (though acknowledge both are possible!). Has anybody actually tested and shown that including the `sound` property in its original location does _not_ work? In other words, are we confident this isn't just a typo in the example?

pushy/src/main/java/com/eatthepath/pushy/apns/PushType.java
```diff
@@ -139,7 +139,22 @@ public enum PushType {
...
+     * documentation:</p>
+     *
+     * <blockquote><p>Set the apns-topic header field of the request you sent to APNs using the following format:
+     * {@code <your bundleID>.push-type.liveactivity.}</p>
+     *
+     * <p>To update a Live Activity, set the value for the payload’s event key to update. To end a Live Activity,
+     * set it to end. If you end a Live Activity, include the final content state to make sure the
+     * Live Activity displays the latest data after it ends.</p></blockquote>
+     *
+     * @since 0.15.2
```
jchambers, line 155: I respect that you're following the pattern we set in the Javadoc for other fields in this enumeration, but I don't think the pattern quite fits in this case. For everything else, we're able to quote directly from https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns to explain the _purpose_ of each push type. Frustratingly, Live Activity push notifications still aren't documented there, so we can't quote it directly. Beyond that, the quote here doesn't really explain what this push type is for. It _does_ explain other important details about topics ad payloads, but I don't think this is (structurally!) where I'd expect to find those details.

So, in short, I think there are a few issues here:

1. We can't directly quote a parallel explanation for what this `push-type` is for
2. We're actually citing a _different_ set of documentation
3. There are other details in here that don't fit the structural pattern established for other keys

So with all that in mind, I'd propose some rephrasing:

```suggestion
     * Indicates that a push notification is intended to update a running Live Activity. Note that Live Activity updates
     * must be sent to a specific topic ({@code [base bundle ID].push-type.liveactivity}) and include specific payload
     * keys (see [insert links here when we sort out what's happening with payload builders]).
     *
     * @since 0.15.2
     *
     * @see <a href="https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications">
     *     Updating and ending your Live Activity with remote push notifications</a>

```

pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityEvent.java
```diff
@@ -0,0 +1,17 @@
+package com.eatthepath.pushy.apns.util;
+
+public enum LiveActivityEvent {
```
jchambers, line 3: I'm cool with enumerating these, but I do predict that, since this is a new feature, there will be some iteration on what's allowed in here in the not-too-distant future. Having this in an enum means that we need to put out a new release of Pushy if Apple ever changes the allowed values.

We _could_ think about doing what we do with `ApnsPayloadBuilder#DEFAULT_SOUND_FILENAME` and just making these string constants (`LIVE_ACTIVITY_EVENT_UPDATE` and `LIVE_ACTIVITY_EVENT_END`) for example, then letting callers set the event type as a string. That would make things a little more flexible with all the good and bad that comes with flexibility.

On balance, though, I think enumerating is probably the right move. That said, if we're introducing a new enum, I think it needs some Javadoc.

pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityApnsPayloadBuilder.java
```diff
@@ -0,0 +1,327 @@
...
+        this.contentState.put(key, value);
+        return this;
+    }
+
+    /**
+     * <p>Sets whether the notification payload will be used for updating the Live Activity
+     * or for ending it.</p>
+     *
+     * @param event {@code LiveActivityEvent.UPDATE} to update the Live Activity or
+     * {@code LiveActivityEvent.END} to end it.
```
jchambers, line 197: Since this is an enumeration, I think it's okay to be a little less verbose here and let the documentation on `LiveActivityEvent` (that we're definitely going to add, riiiiiight?) explain the options in more detail:

```suggestion
     * @param event the type of update event to send to the destination activity
```

pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityApnsPayloadBuilder.java
```diff
@@ -0,0 +1,327 @@
...
+     *
+     * @param key the key of the custom property in the content-state object
+     * @param value the value of the custom property
+     *
+     * @return a reference to this payload builder
+     *
+     * @see <a href="https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications">
+     *      Updating and ending your Live Activity with remote push notifications</a>
+     */
+    public LiveActivityApnsPayloadBuilder addContentStateProperty(final String key, final Object value) {
```
jchambers, line 187: Rather than specifying `content-state` on key at a time, why not just let callers provide a `Map<String, Object>`?

```suggestion
    public LiveActivityApnsPayloadBuilder setContentState(final Map<String, Object> contentState) {
```

I recognize that there's some precedent from `ApnsPayloadBuilder#addCustomProperty`, but we need to be working with individual keys there because we're asking callers to add to a map that has competing "owners;" we can't let callers clobber the `aps` property, for example. By contrast, callers "own" all of the `content-state` map and can just specify the whole thing in a single shot.

jchambers (changes requested):
First and foremost, thank you for this contribution and for your patience!

I have a few (relatively minor!) documentation requests, but the main thing I'm skeptical of is adding _an entirely new kind of payload builder._ As I mentioned in a [separate comment](https://github.com/jchambers/pushy/pull/990#issuecomment-1304565240), I'm skeptical that this is actually an intentional change on Apple's part, and I think we need concrete test results showing that (a) the proposed new topology _does_ work _and_ (b) the old topology _does not_ work before we make such a major change to Pushy.

Even if it turns out that we do need to move the `sound` key for Live Activity updates, I'd much rather just add a flag to the existing payload builder (or just add some conditional logic based on the presence of an `event`/`content-state` value); please see `ApnsPayloadBuilder#preferStringRepresentationForAlerts` for precedent.

ifrins:
Hey @jchambers I did test the sound property, and it's a bit of a strange situation:

This produces a sound when delivered:
```json
{
	"aps": {
		"alert": {
			"body": "Body",
			"title": "Title"
		},
		"content-state": {},
		"event": "update",
		"timestamp": 1667664660
	}
}
```

This does not:
```json 
{
	"aps": {
		"alert": {
			"body": "Body",
			"title": "Title",
			"sound": ""
		},
		"content-state": {},
		"event": "update",
		"timestamp": 1667664660
	}
}
```

and I haven't really been able to make it use a custom sound but 🤷. Sound vs. silent does not work if it's in `$.aps`, so this doesn't do anything (aka. it makes sound):

```json
{
	"aps": {
		"sound": "",
		"alert": {
			"body": "Body",
			"title": "Title"
		},
		"content-state": {},
		"event": "update",
		"timestamp": 1667664834
	}
}
```
And FWIW I'm relying on using sound property in production to silence the alerts. The behavior is the same on iOS 16.1 and 16.2 beta. 

--- 

About having a separate PayloadBuilder vs not: I personally lean towards separate Payload Builders. 

Live Activity alert payloads are very _special_. For example, the alert is only shown if the tile and body are present. The rest of the usual `$.aps.alert` properties are completely ignored so it looks a bit confusing. I personally think having a PayloadBuilder that only accepts properties applicable to Live Activities is helpful when creating the payload (also applicable for the other push types that are not related to sending an alert).

 But that, in the end, it's also a matter of preference and if you think having everything on the same PayloadBuilder is the preferred direction for the project, then I'll move it back :)

jchambers:
> I did test the sound property, and it's a bit of a strange situation…

That is, indeed, strange. I'm sorry to impose, but can you confirm that this behavior is different from what happens if you're sending a "conventional" push notification? In other words, can we be confident that this is tied specifically to Live Activity updates as opposed to being an undocumented change in iOS 16 in general?

> …if you think having everything on the same PayloadBuilder is the preferred direction for the project…

I actually wouldn't go that far; I think the current system is already showing signs of strain and it's worth thinking about an overhaul. For now, though, I think it makes sense to keep the surface area small as possible and just add new properties to the existing payload builder. Let's definitely bear this in mind for future revision, though.

One thing that makes me want to be a little more careful about large changes to the payload builder system is that it _feels_ like something is changing upstream, but it's still not well-documented. I think I want to understand where things are going a little better before committing to a new design approach.

Again, thanks for the thoughtful work on all this!

ifrins:
Yup I also checked the behavior for regular alert pushes. Setting a sound in `$.aps.sound` does nothing. 

This doesn't work in regular alert pushes:
```json
{
	"aps": {
		"alert": {
			"body": "Body",
			"sound": "default"
		}
	}
}
```

and, as expected, this makes sound:
```json
{
	"aps": {
		"alert": {
			"body": "Body"
		},
		"sound": "default"
	}
}
```

so at least as of iOS 16.1 and 16.2 the usage of `$.aps.alert.sound` seems limited to alerts triggered by Live Activities.

jchambers:
> Yup I also checked the behavior for regular alert pushes. Setting a sound in `$.aps.sound` does nothing.

Thank you for checking! Super weird that they'd make it different just for this use case, but that seems to be where the evidence is pointing.

jchambers:
@ifrins No rush, but I did just want to check to make sure we're on the same page—are you planning on making the requested changes? I'm happy to cook something up if not! I just want to make sure we're not getting into a situation where each of us expects the other to be doing something we weren't planning on doing!

ifrins:
Hey, sorry! I haven't had a chance to address the changes yet. I will probably have some time to do it this upcoming week.

jchambers:
No problem at all! I will be the _last_ person to criticize anybody else for being slow on this pull request ;)

Like I said: I just wanted to make sure we had a shared understanding of who's doing what. I'll look forward to reviewing the changes when they're available!

ifrins:
Apologies for the delay! I think I have addressed all the comments in the review @jchambers

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -715,6 +729,76 @@ public ApnsPayloadBuilder addCustomProperty(final String key, final Object value
         return this;
     }
 
+    /**
+     * <p>Sets the provided content state object inside content-state payload for Live Activities.</p>
```
jchambers, line 733: Trivial (and reasonable minds might disagree):

```suggestion
     * <p>Sets the content state object inside content-state payload for Live Activities.</p>
```

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -715,6 +729,76 @@ public ApnsPayloadBuilder addCustomProperty(final String key, final Object value
...
+     * @see <a href="https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications">
+     *      Updating and ending your Live Activity with remote push notifications</a>
+     */
+    public ApnsPayloadBuilder setEvent(final LiveActivityEvent event) {
+        this.event = event;
+        return this;
+    }
+
+    /**
+     * <p>Sets a timestamp for the push notification payload.   </p>
```
jchambers, line 767: Trivial:

```suggestion
     * <p>Sets a timestamp for the push notification payload.</p>
```

Also, I'm sorry I'm only noticing this now, but other than appearing in the example payload at https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications, I can't find any mention of the `timestamp` field in either the server-side or client-side docs; am I missing something? Do we know definitively what this is for or what happens if it gets omitted?

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -849,6 +953,10 @@ protected Map<String, Object> buildPayloadMap() {
         return payload;
     }
 
+    private boolean isLiveActivityPayload() {
+        return event != null;
+    }
```
jchambers, line 958: Trivial: since this is only used in one method and since it's not being exposed for tests, I'd be tempted to just inline this in `buildPayloadMap` as:

```
final boolean isLiveActivityPayload = event != null;
```

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -833,6 +933,10 @@ protected Map<String, Object> buildPayloadMap() {
                 if (this.launchImageFileName != null) {
                     alert.put(LAUNCH_IMAGE_KEY, this.launchImageFileName);
                 }
+
+                if (this.soundFileName != null && isLiveActivityPayload()) {
+                    alert.put(SOUND_KEY, this.soundFileName);
+                }
```
jchambers, line 939: Minor: I'd say this is worth a comment explaining the strange payload situation we discussed earlier. Without a  comment, my fear is that somebody (i.e. me in six months) will look at this and think "well that's strange," then try to "fix" it.

pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityEvent.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    END("end");
+
+    private final String value;
+
+    LiveActivityEvent(final String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
```
jchambers, line 49: Trivial: I think this could be package-private, right?

pushy/src/test/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilderTest.java
```diff
@@ -611,6 +612,82 @@ void testBuildMdmPayload() {
...
+        final Instant dismissalDate = Instant.now();
+        this.builder.setDismissalDate(dismissalDate);
+
+        final Map<String, Object> aps = this.extractApsObjectFromPayloadString(this.builder.build());
+
+        assertEquals(dismissalDate.getEpochSecond(), aps.get("dismissal-date"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("argumentsForSetEvent")
```
jchambers, line 649: Minor: I think you could simplify this by using [`@EnumSource`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-EnumSource) instead of `@MethodSource`.

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -715,6 +729,76 @@ public ApnsPayloadBuilder addCustomProperty(final String key, final Object value
...
+     */
+    public ApnsPayloadBuilder setEvent(final LiveActivityEvent event) {
+        this.event = event;
+        return this;
+    }
+
+    /**
+     * <p>Sets a timestamp for the push notification payload.   </p>
+     *
+     * @param timestamp Timestamp
```
jchambers, line 769: Can we say more about this? I guess this is related to the other question, though, in that I guess we need to find some explanation for what this really does before we can explain it to Pushy's end users.

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -715,6 +729,76 @@ public ApnsPayloadBuilder addCustomProperty(final String key, final Object value
...
+    /**
+     * <p>Sets a timestamp for the push notification payload. The timestamp is used to discard older
+     * push notifications. According to Apple's documentation:</p>
+     *
+     * <blockquote>When you end a Live Activity, by default the Live Activity appears on the Lock Screen for up to
+     * four hours after it ends to allow the user to glance at their phone to see the latest information. To dismiss
+     * the Live Activity from the Lock Screen immediately after it ends, provide a date for "dismissal-date" that’s
+     * in the past. Alternatively, provide a date within a four-hour window to set a custom dismissal date.</blockquote>
+     *
+     * @param dismissalDate Date when the Live Activity will be dismissed
```
jchambers, line 790: Minor: teeeeeechnically this isn't a "date" by `java.time` conventions (i.e. a "date" would be something like "November 27, 2022").

```suggestion
     * @param dismissalDate the instant when the Live Activity will be dismissed
```

jchambers (approved):
This is looking really good to me! Thank you for all of your work on this!

I have a smattering of small suggestions, but those could reasonably be ignored (and/or I could just do a small follow-up commit after merging this). The one thing I would like to address before merging this: do we know what's up with the "timestamp" field? Are we sure that's a real thing and not just a weird artifact in the example payload?

ifrins:
@jchambers Thanks for the review!

About the `timestamp` field, there's no _public_ info I could find on the property. From what I have seen while building an app that uses Live Activities, this is being used by the system for versioning. The Live Activity only updates if `new timestamp > old timestamp` otherwise it discards the update. 

I didn't add to the javadoc mainly because I couldn't find an authoritative source on what it does. 

I agree on all the other suggestions, will update the PR.

jchambers:
> About the timestamp field, there's no _public_ info I could find on the property. From what I have seen while building an app that uses Live Activities, this is being used by the system for versioning. The Live Activity only updates if new timestamp > old timestamp otherwise it discards the update.

That makes sense (though it's frustrating that there's no documentation). Thanks for sharing your practical experience—that helps a lot!

> I didn't add to the javadoc mainly because I couldn't find an authoritative source on what it does.

I get that, but I do think we should put _something_ there. The current text is:

```
Sets a timestamp for the push notification payload.

@param timestamp Timestamp
```

I'd propose at least calling out that this is associated with Live Activities, and then we can at least offer our best guess as to what callers ought to put there. So maybe something like:

```
Sets the Live Activity timestamp for the push notification payload.

@param timestamp the instant at which the Live Activity event occurred
```

…but once that's settled, I look forward to merging this!

jchambers:
Once again, thank you for the thoughtful contribution! I'll merge this as soon as tests pass and will pull together a release shortly thereafter. To manage expectations, getting the release done might happen a little later in the week, but it'll happen soon.

PR #1000, opened 2022-12-07
support push to talk push notifications

lichongbing:
support push to talk push notifications


jchambers:
@lichongbing Again, thank you for the contribution. There are still several things here that aren't quite right (again, please don't increment version numbers in `pom.xml`), and I think the easiest way forward will be for me to just go ahead and make the changes myself. To that end, I've opened #1005 and will merge that shortly.

Thank you very much for your work and for this contribution!

PR #1001, opened 2022-12-09
Added HTTP/2-ALPN to SslContextBuilder in ApnsClientBuilder.build()

MarcelHeckel:
The pushy client was not working, when an IBM-ALG (application layer gateway) was set as revers proxy for the Apple server, because the ALG did not support HTTP/2 via Prior Knowledge. It must be negotiated via TLS.

The HTTP/2 standard says (https://www.rfc-editor.org/rfc/rfc7540 - section 3.4.): "implementations that support HTTP/2 over TLS MUST use protocol negotiation in TLS [TLS-ALPN]."

The added code for ALPN (application_layer_protocol_negotiation) is bases on:
* https://netty.io/wiki/requirements-for-4.x.html
* https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/http2/helloworld/client/Http2Client.java#L87

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java
@@ -85,6 +85,11 @@ public class ApnsClientBuilder {
     private Http2FrameLogger frameLogger;
 
     /**
+     * @see #setUseAlpn
+     */
+    private boolean useAlpn = false;
+
+    /**
      * The default idle time in milliseconds after which the client will send a PING frame to the APNs server.
      *
      * @since 0.11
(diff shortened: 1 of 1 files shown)


jchambers:
Thank you for the contribution! I understand why you want this, but have some reservations based on past experience.

The earliest versions of Pushy that supported HTTP/2 actually used ALPN, and it was a nightmare. JDK-level support for ALPN was not widespread, and the challenges of using an external provider (either netty-tcnative or jetty-alpn-agent) caused no shortage of problems. At the time, this was a problem because Pushy targeted Java 7, and Java 7 had no built-in support for ALPN.

My understanding (which could be wrong!) is that ALPN support is available in Java 8 (which Pushy now targets), but only after a certain update. If it's the case that some sub-versions of Java 8 support ALPN and some don't, this change risks creating a situation where Pushy mysteriously stops working for folks using a supposedly-supported version of Java.

To move this forward, I'll do some research to understand precisely when ALPN support got added to Java 8. If it is, indeed, a case where we'd be breaking Pushy for existing users, then I think we'd be making life worse for more people than we'd be helping.

The way forward here might be to make ALPN support optional and have it default to "off." We could add something like `ApnsClientBuilder#setUseAlpn(boolean)` for the exceptionally small slice of users for whom this is an issue. For context, we already do something similar with Pushy's mock server: https://github.com/jchambers/pushy/blob/5862c537899381b44ec3e8b0b904fc8e07740ba1/pushy/src/main/java/com/eatthepath/pushy/apns/server/BaseHttp2ServerBuilder.java#L323-L329

MarcelHeckel:
> The way forward here might be to make ALPN support optional and have it default to "off." We could add something like ApnsClientBuilder#setUseAlpn(boolean) for the exceptionally small slice of users for whom this is an issue. For context, we already do something similar with Pushy's mock server:

That would be OK for me! Should I update the PR?

jchambers:
> Should I update the PR?

@MarcelHeckel I'm extremely sorry; I missed your comment ages ago. Yes, if you'd like to, that would be a welcome contribution!

MarcelHeckel:
@jchambers I have updated the PR with the config option.

jchambers (approved):
I'm embarrassed that it's taken me nearly a year to follow up on this. Thank you both for your patience and for this contribution!

PR #1024, opened 2023-05-02
Update to Netty 4.1.92

roberttaylor426:
Patches vulnerability [CVE-2022-41881](https://nvd.nist.gov/vuln/detail/CVE-2022-41881).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,7 +154,7 @@
     </dependencyManagement>
 
     <properties>
-        <netty.version>4.1.85.Final</netty.version>
+        <netty.version>4.1.92.Final</netty.version>
         <junit.version>5.6.0</junit.version>
         <slf4j.version>1.7.21</slf4j.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
--- a/pom.xml
+++ b/pom.xml
@@ -105,7 +105,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-tcnative-boringssl-static</artifactId>
-                <version>2.0.54.Final</version>
+                <version>2.0.60.Final</version>
             </dependency>
             <dependency>
                 <groupId>com.eatthepath</groupId>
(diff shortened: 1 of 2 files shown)


jchambers (changes requested):
Hello, and thank you for the contribution!

I think this is a good change in general, but this is not an urgent security issue because it is specific to `io.netty.codec:codec-haproxy`, which Pushy does not use. With that in mind, I think it makes sense to merge this, but not to trigger a new release of Pushy just for this change.

Additionally, it looks like the release of Netty 4.1.92 is coupled with the release of netty-tcnative 2.0.60; I think we need to update that, too, as part of this change.

roberttaylor426:
> it looks like the release of Netty 4.1.92 is coupled with the release of netty-tcnative 2.0.60; I think we need to update that, too, as part of this change.

Done.

> I think it makes sense to merge this, but not to trigger a new release of Pushy just for this change

I'd be grateful if you could cut a new release, as your library is currently failing an OWASP dependency check.

pom.xml
```diff
@@ -105,7 +105,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-tcnative-boringssl-static</artifactId>
-                <version>2.0.54.Final</version>
+                <version>2.0.60.Final</version>
```
jchambers, line 108: I hate to be That Guy, but I think there's a reference in the README that needs to get updated, too.

jchambers:
> …as your library is currently failing an OWASP dependency check.

Respectfully, that seems pretty strange to me; Pushy doesn't depend on the affected artifact. Can you please be more specific about what tool you're using and which dependency is getting flagged?

roberttaylor426:
Ok that's fair. 

The tool I'm using is [DependencyCheck](https://github.com/jeremylong/DependencyCheck). I guess the reason it's flagging your library is because the CPE in the [CVE](https://nvd.nist.gov/vuln/detail/CVE-2022-41881) in the NVD (phew!) is too broad: `cpe:2.3:a:netty:netty:*:*:*:*:*:*:*:*`.

jchambers (approved):

jchambers:
> …the CPE in the [CVE](https://nvd.nist.gov/vuln/detail/CVE-2022-41881) in the NVD (phew!) is too broad.

Ah—yes, it looks like you're right. I'm still not exactly excited about putting out a release that takes on the risk of a major dependency update with no clear payoff and no other changes, but I'll consider it. Thanks again for the contribution!

andreoss:
@jchambers @roberttaylor426 
Would it be possible to update to 4.1.94.Final, as this version also addresses CVE-2023-34462?

jchambers:
Thank you for this contribution. I've updated to the latest Netty 4.1.104 and netty-tcnative 2.0.62 in a63e07c.

PR #1040, opened 2023-10-06
Simplify authentication token encoding/decoding and verification

jchambers:
This pull request does two things:

1. It remedies potential signature verification errors (described in #1037) in `AuthenticationToken` and
2. It moves away from Netty's more complicated Base64 codec to the much more compact JDK Base64 codec


KiruthikaJanakiraman:
I tested the fix as well with [NonDex](https://github.com/TestingResearchIllinois/NonDex) and the flaky tests pass.

bbelide2:
@jchambers Just wanted to know about the status of this PR. Will this change fix the issue ? If Yes, is this change ready to be merged?

jchambers:
@bbelide2 I don't know that we're on the same page here. Is this actually causing a problem for you?

jchambers:
Frankly, I have a long backlog of Pushy stuff to resolve, and this is part of it. I'll probably try to plow through all of it over the winter break in a few weeks.

PR #1043, opened 2023-10-21
Support apns unique id

hectorespert:
Read 'apns-unique-id' from response headers. Closes https://github.com/jchambers/pushy/issues/1028

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
@@ -360,23 +360,20 @@ class ApnsClientHandler extends Http2ConnectionHandler implements Http2FrameList
     }
 
     private static UUID getApnsIdFromHeaders(final Http2Headers headers) {
-        final CharSequence apnsIdSequence = headers.get(APNS_ID_HEADER);
-
-        try {
-            return apnsIdSequence != null ? FastUUID.parseUUID(apnsIdSequence) : null;
-        } catch (final IllegalArgumentException e) {
-            log.error("Failed to parse `apns-id` header: {}", apnsIdSequence, e);
-            return null;
-        }
+        return getUUIDFromHeaders(headers, APNS_ID_HEADER);
     }
 
     private static UUID getApnsUniqueIdFromHeaders(final Http2Headers headers) {
-        final CharSequence apnsIdSequence = headers.get(APNS_UNIQUE_ID_HEADER);
+        return getUUIDFromHeaders(headers, APNS_UNIQUE_ID_HEADER);
+    }
+
+    private static UUID getUUIDFromHeaders(final Http2Headers headers, final AsciiString header) {
+        final CharSequence apnsIdSequence = headers.get(header);
 
         try {
             return apnsIdSequence != null ? FastUUID.parseUUID(apnsIdSequence) : null;
         } catch (final IllegalArgumentException e) {
-            log.error("Failed to parse `{}` header: {}", APNS_UNIQUE_ID_HEADER, apnsIdSequence, e);
+            log.error("Failed to parse `{}` header: {}", header, apnsIdSequence, e);
             return null;
         }
     }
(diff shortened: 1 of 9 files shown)


pushy/src/main/java/com/eatthepath/pushy/apns/PushNotificationResponse.java
```diff
@@ -62,6 +62,16 @@ public interface PushNotificationResponse<T extends ApnsPushNotification> {
...
 
+
+    /**
+     * Returns a unique ID only available in the development environment.
+     * Useful to query push information in Push Notifications Console.
+     * @return UUID The apns unique id
+     */
+    default Optional<UUID> getApnsUniqueId() {
+        return Optional.empty();
+    }
```
jchambers, line 73: I'm not opposed to using `default` methods in general, but this would be (I believe) the first and only use in Pushy. For now, I'd recommend leaving this without a default implementation and revisiting the issue of default implementations more holistically in a separate effort/pull request.

pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientHandler.java
```diff
@@ -352,17 +354,26 @@ protected void handleErrorResponse(final ChannelHandlerContext context, final in
...
-        final CharSequence apnsIdSequence = headers.get(APNS_ID_HEADER);
+        return getUUIDFromHeaders(headers, APNS_ID_HEADER);
+    }
+
+    private static UUID getApnsUniqueIdFromHeaders(final Http2Headers headers) {
+        return getUUIDFromHeaders(headers, APNS_UNIQUE_ID_HEADER);
+    }
+
+    private static UUID getUUIDFromHeaders(final Http2Headers headers, final AsciiString header) {
+        final CharSequence apnsIdSequence = headers.get(header);
```
jchambers, line 371: I think generalizing this method makes sense, but I think we can also go one step further and generalize this variable's name.

```suggestion
        final CharSequence uuidSequence = headers.get(header);
```

pushy/src/main/java/com/eatthepath/pushy/apns/PushNotificationResponse.java
```diff
@@ -62,6 +62,16 @@ public interface PushNotificationResponse<T extends ApnsPushNotification> {
      */
     UUID getApnsId();
 
+
+    /**
+     * Returns a unique ID only available in the development environment.
+     * Useful to query push information in Push Notifications Console.
+     * @return UUID The apns unique id
+     */
```
jchambers, line 70: I think this is good, but if you'll forgive some nitpicking, I think we can be more specific and more explicitly cross-reference this with other parts of our own documentation and Apple's documentation.

```suggestion
    /**
     * Returns a unique identifier set by the APNs server in development environments to facilitate testing
     * notifications. Note that this identifier is distinct from the identifier returned by {@link #getApnsId()}.
     *
     * @return the unique identifier assigned by the APNs server to the sent notification if the server is in the APNs
     * development environment or empty if the server is in the production environment and did not include a unique
     * identifier
     *
     * @see ApnsClientBuilder#setApnsServer(String)
     * @see ApnsClientBuilder#DEVELOPMENT_APNS_HOST
     * @see <a href="https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns#3394541">Handling notification responses from APNs - Interpret header responses</a>
     * @see <a href="https://developer.apple.com/documentation/usernotifications/testing_notifications_using_the_push_notification_console">Testing notifications using the Push Notification Console</a>
     */
```

jchambers (commented):
Thank you for this contribution! I'm very sorry it's taken me so long to give it the attention it deserves. I think this might benefit from a few small adjustments. If you'd like to update this pull request, please do! If not, I'm comfortable merging it as-is and will make the adjustments in a follow-up commit.

pushy/src/main/java/com/eatthepath/pushy/apns/PushNotificationResponse.java
jchambers: Ah—but then again, this will preserve backward compatibility for existing, third-party `PushNotificationResponse` implementations. Okay. I think I'm convinced.

jchambers:
@hectorespert Thinking about this a little more, I'll just go ahead and merge this and do the fiddly cleanup stuff on my own in a follow-up commit. Once again, thank you for this contribution!

PR #1046, opened 2023-12-31
Significantly overhaul the `ApnsClientMetricsListener` interface

jchambers:
This (in my opinion) greatly improves the `ApnsClientMetricsListener` interface at the cost of making a bunch of breaking changes. The key changes are:

1. We no longer issue opaque "notification IDs" to metrics listeners, but instead provide "send notification" durations where appropriate
2. We now consolidate accepted/rejected handlers into a single "acknowledged" handler that has access to the full notification response
3. The Micrometer implementation is now much, much more sensibly constructed; when I first built it, I had never really used Micrometer before and was mostly guessing at how it worked. Now that I have significant direct experience with it, I can offer a much more sensible and ergonomic design.

Feedback on all of this is very welcome.


jhonynet:
Hi Jon! do you have any plans to release this soon?

jchambers:
No specific plan other than "next." With WWDC happening, though, I expect there will be at least _some_ changes that demand a more immediate response, and that's likely to prompt a more concentrated burst of activity on this project from me.

PR #1049, opened 2024-01-15
Add a "start" event for Live Activities

### LINKED ISSUE #1047: Missing "start" enumeration com.eatthepath.pushy.apns.util.LiveActivityEvent
The official apns documentation requires the event to be specified as start, but pushy does have this enumerated value.

https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications#Construct-the-payload-that-starts-a-Live-Activity

What do you think about it?

jchambers:
Evidently, the process for starting Live Activities has changed in the latest iOS release. The old documentation (https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications) appears to have been replaced with https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications, which now calls for explicit "start" events and makes no mention of "end" events.

It's unclear to me if this change is a replacement for the old behavior, an optional replacement, or a mandatory replacement. In any case, I think the safest thing to do is simply to add the new "start" event.

This fixes #1047.

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityEvent.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/util/LiveActivityEvent.java
@@ -25,11 +25,16 @@ package com.eatthepath.pushy.apns.util;
 /**
  * An enumeration of accepted actions for Live Activity notification payloads.
  *
- * @see <a href="https://developer.apple.com/documentation/activitykit/update-and-end-your-live-activity-with-remote-push-notifications">Updating and ending your Live Activity with remote push notifications</a>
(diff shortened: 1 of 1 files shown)


pushly-cj:
Just confirming that the "start" event is a new event introduced in iOS 17.2 that allows a Live Activity to be started via an APNS Live Activity Notification whereas before you could only start the activity via your app and perform "update" events via APNS. I.e., The "update" event is still relevant and separate from "start". 

We are seeing that the "end" event is still supported via APNS regardless of it not being in the documentation.

jchambers:
@pushly-cj Thank you—this is very helpful!

PR #1051, opened 2024-01-28
Live activity attributes type

### LINKED ISSUE #1050: add attributes-type and attributes to the payloadbuilder
In the [documentation for remote starting a live activity](https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications#Construct-the-payload-that-starts-a-Live-Activity) it is mentioned that the attributes and attributes-type keys should be present in the payload when starting a live activity. These are not available in the payload builder yet.

jchambers:
This adds support for the `attributes-type` and `attributes` payload fields for Live Activities.

This fixes #1050.


erickok:
Thanks! This looks good. We are testing remote live activity starting and are using your library. Any release planned?

jchambers:
I'll put out a release when I can get a few contiguous free hours to do so. That said, I'm not treating this as an emergency because it only affects payload construction, and callers are free to create payload strings however they'd like; there's no requirement that you actually use Pushy's included payload builder!

Issue #617, opened 2018-06-19
Use Travis CI deployments to manage Pushy releases


jchambers:
Right now, the deployment process for Pushy is that I type a bunch of stuff into a terminal to manually trigger deployment to Maven Central using the [Maven deploy plugin](http://maven.apache.org/plugins/maven-deploy-plugin/). This is more error-prone than is ideal, and also perhaps excessively dependent on my specific credentials.

As an alternative, we should consider using the [Travis CI deployment](https://docs.travis-ci.com/user/deployment) system to build, sign, and publish artifacts at release time.

jchambers:
Now that [GitHub Actions](https://github.com/actions) are A Thing, we might consider using those, too.

PR #1075, opened 2024-07-06
Use `jgitver` to manage versions

jchambers:
In support of #617, this change introduces [jgitver](https://github.com/jgitver/jgitver) to set Maven versions based on git tags. This should make some parts of automated builds/releases much easier.


jchambers:
I'm torn about whether this is actually a good idea. On one hand, it'll make automated builds much easier. On the other, it introduces some new stumbling blocks, especially for IntelliJ users (explained in the README!). Feedback on this change is very welcome!

PR #1079, opened 2024-08-17
Add support for starting a Live Activity with input-push-token on iOS 18

ifrins:
Adds support for passing the `input-push-token` attribute, which is required in iOS 18 to start a Live Activity that wakes the app to get a push token to update the created Live Activity (which was not needed on iOS 17).

This has been tested with iOS 18.0 beta 6. 

Apple docs: https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
@@ -859,8 +859,12 @@ public abstract class ApnsPayloadBuilder {
     }
 
     /**
-     * <p>Sets the flag to wake up the app and receive a push token to send updates to the started Live Activity.</p>
-     *
+     * Sets the flag to control whether the OS wakes up the app and delivers a push token to send updates to the
+     * started Live Activity. It is only respected in iOS 18 and later. Older versions provide a new push token
+     * by default.
+
+     * @param inputPushToken Flag to decide whether the app is provided a push token to update the Live Activity
+
      * @return a reference to this payload builder
      *
      * @see <a href="https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications">
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
@@ -82,6 +82,7 @@ public abstract class ApnsPayloadBuilder {
     private boolean preferStringRepresentationForAlerts = false;
 
     private LiveActivityEvent event = null;
+    private Integer inputPushToken = null;
 
     private Instant timestamp = null;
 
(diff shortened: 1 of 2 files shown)


pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -856,6 +858,19 @@ public ApnsPayloadBuilder setTimestamp(final Instant timestamp) {
...
 
+    /**
+     * <p>Sets the flag to wake up the app and receive a push token to send updates to the started Live Activity.</p>
+     *
+     * @return a reference to this payload builder
+     *
+     * @see <a href="https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications">
+     *      Starting and updating Live Activities with ActivityKit push notifications</a>
+     */
+    public ApnsPayloadBuilder setInputPushToken() {
```
jchambers, line 869: We have, so far, avoided adding "one-way" switches to `ApnsPayloadBuilder`; if callers can add a thing to a payload, they ought to be able to remove it, too. My recommendation would be to model this as a boolean:

```suggestion
    public ApnsPayloadBuilder setInputPushToken(final boolean inputPushToken) {
```

…and then callers can change their minds about the property later.

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
```diff
@@ -856,6 +858,19 @@ public ApnsPayloadBuilder setTimestamp(final Instant timestamp) {
         return this;
     }
 
+    /**
+     * <p>Sets the flag to wake up the app and receive a push token to send updates to the started Live Activity.</p>
```
jchambers, line 862: A couple comments here:

1. If this is a single paragraph, I don't think we need `<p>` tags.
2. While this isn't incorrect, I do think there's more of the story we can and should tell for callers. In particular: this is for establishing a new channel for live activities, and it's intended for iOS 18 and newer. If it's helpful to you, I'd be happy to suggest specific wording!

jchambers (commented):
Hello, and thank you for the contribution! It's unclear to me why Apple has decided to model this as an integer; it seems like it's a boolean flag, right?

I think it probably makes sense to change the internal representation from a nullable `Integer` to just a `boolean`, then turn that into a `1` when we actually build the payload. Does that seem reasonable to you?

pushy/src/main/java/com/eatthepath/pushy/apns/util/ApnsPayloadBuilder.java
ifrins: Sorry for the delay! I have updated the Javadoc to provide a bit more info. If the changes are fine, otherwise I'm open for any suggestions. In general, I don't like rehashing info from the source (Apple docs in this case), since from the outside most of the time it feels like trying to documenting undocumented behavior.

Clon1998:
Any update on this PR?

jchambers:
@ifrins With apologies for the delay (and as I mentioned separately in #1097), there's enough going on in the live activity/broadcast notification space that I think it's probably going to make more sense for me to just tackle the whole thing holistically.

I really really appreciate this contribution, but am going to decline for now in favor of a larger (and immediate!) effort to just do the whole broadcast notification system in a single shot.

PR #1086, opened 2024-10-28
Custom SslContext option added to ApnsClientBuilder

kkolyan:
1. we use a multiple APNS clients with different credentials
2. we periodically create new one-shot clients to send a single notifications in order to check credentials.

du to the facts mentioned above, SslContext initialisation takes considerable amount of CPU. The goal of this change is to reduce it by sharing SslContext between clients (we use signingKey to authenticate clients, so seems like it's correct to share SslContext).

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java
@@ -85,6 +85,8 @@ public class ApnsClientBuilder {
 
     private boolean useAlpn = false;
 
+    private SslContext sslContext;
+
     /**
      * The default idle time after which the client will close a connection (which may be reopened later).
      *
@@ -577,6 +579,26 @@ public class ApnsClientBuilder {
     }
 
     /**
+     * <p>Sets custom SslContext instance.</p>
+     * <p>this property is mutually exclusive with SslContext is set using Overrides effect of {@link #trustedServerCertificatePemFile}, {@link #trustedServerCertificateInputStream},
+     * {@link #trustedServerCertificates}, {@link #clientCertificate}, {@link #privateKey}, {@link #privateKeyPassword}
+     * and {@link #useAlpn}</p>
+     *
+     * @param sslContext fully initialized SslContext or null.
+     * @return a reference to this builder
+     */
+    public ApnsClientBuilder setSslContext(SslContext sslContext) {
+        if (this.sslContext instanceof ReferenceCounted) {
+            ((ReferenceCounted) this.sslContext).release();
+        }
+        this.sslContext = sslContext;
+        if (sslContext instanceof ReferenceCounted) {
+            ((ReferenceCounted) sslContext).retain();
+        }
+        return this;
+    }
+
+    /**
      * Constructs a new {@link ApnsClient} with the previously-set configuration.
      *
      * @return a new ApnsClient instance with the previously-set configuration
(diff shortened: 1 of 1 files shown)


jchambers:
Hello, and thank you for the contribution!

I don't think I have a clear understanding of your use case. How many `ApnsClient` instances are you building that `SslContext` construction is a significant part of your CPU load?

jchambers:
Thanks for the explanation! That does help me understand the problem a little better, but frankly, something's not adding up for me. You wrote:

> And that activity (~100 checks per minute) takes about 0.35 CPU core in average continuously, which is almost the same as each our node consumes to send normal notifications.

It sounds like you think this time is dominated by constructed `SslContext` instances. For the sake of argument, let's assume that _all_ of the CPU time is going to `SslContext` initialization. If that were true, I think that would mean that we could construct about 300 `SslContext` instances per minute (or 5/second) while completely saturating a CPU core. That seems… extremely slow to me. I wrote a quick benchmark to check how long it takes to construct an `SslContext` the same way we'd build one if we were using a signing key:

```java
package com.eatthepath.pushy.apns;

import io.netty.handler.codec.http2.Http2SecurityUtil;
import io.netty.handler.ssl.*;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;

import javax.net.ssl.SSLException;

@State(Scope.Thread)
public class SslContextBenchmark {

  @Benchmark
  public SslContext buildSslContext() throws SSLException {
    final SslContextBuilder sslContextBuilder = SslContextBuilder.forClient()
        .sslProvider(SslProvider.OPENSSL)
        .ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE);

    return sslContextBuilder.build();
  }
}
```

…and on my humble laptop, I'm generating upwards of 50,000 `SslContext` instances per second on a single core:

```
Benchmark                             Mode  Cnt      Score     Error  Units
SslContextBenchmark.buildSslContext  thrpt   25  52587.568 ± 232.103  ops/s
```

That's four orders of magnitude off from what we'd expect from your numbers. I don't think `SslContext` construction is your main problem, and I'd encourage you to do more detailed profiling to identify the problem.

On top of that, I think it's fair to say that your specific use case is uncommon, and reusing `SslContext` instances may have unexpected consequences for users who are new to the space. With all of those things in mind, I must respectfully decline this contribution. All the same, thank you for your work!

jchambers:
> the connection of my problem with SslContext CPU costs looks doubtful

Well, more precisely, I'd say you're having a problem that's not common or easy to reproduce. From your flamegraph, it looks like Bouncycastle is spending an enormous amount of time… decrypting a trust store, it looks like? Perhaps there's a way to address that problem separately.

It sounds like you have a good understanding of the problem, though:

> we use use FIPS-ed JVM due to customer security requirements, so the set of cryptographic algorithms and its parameters (key lengths for instance) is different from the ones you use in the mentioned benchmark. And probably, the one we use makes SslContext creation much more slower that default ones you used.

But, yes:

> the case is not common enough to address it in mainstream.

…you're correct that this is my main point. It sounds like you have an uncommon use case (spawning lots of single-shot clients) combined with an uncommon problem (a higher-than-usual CPU load when loading trust stores). It's entirely possible that you're the only Pushy user who would benefit from the feature you're requesting, and it would come at a cost of complexity and some potential security risk to other users. For those reasons, I still think it makes sense to leave this change out of Pushy.

I recognize that this may be disappointing for you, but I'm confident it's the right move for Pushy as a whole. I wish you luck in addressing the challenge by other means, though!

Issue #1094, opened 2024-12-27
Memory leak in ApnsChannelPool


Sycamore-M:
Hi，

We have an offline push service. After running for a few months, an out of memory issue occurred.

By analyzing memory snapshots, it was found that there are hundreds of thousands of **_DefaultPromise_** stacked in **_ApnsChannelPool.pendingAcquisitionPromises_**

We had to restart the service regularly before this issue was fixed.

pushy:0.15.4
jdk:11.0.25
netty:4.1.101.Final

Sycamore-M:
I have found that when the SSL certificate expires, ApnsChannelPool.pendingAcquisionPromises will continue to expand.

jchambers:
We discussed this a bit more in #1095. I think we agree that this isn't a memory "leak" per se (all of the memory is still accounted for and recoverable), and this is a thing that can happen with a combination of expired certificates and loose control. Since everything is working as expected within Pushy and since there are external mitigations for this problem (check certificate expiration, limit pending sends), I think it makes sense to close this for now.

Thank you for reporting the issue and for proposing a solution, though!

PR #1095, opened 2024-12-30
Support memory overflow protection.

Sycamore-M:
My attempt to fix: [https://github.com/jchambers/pushy/issues/1094](url)

### DIFF
--- a/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
+++ b/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsChannelPool.java
@@ -62,6 +62,7 @@ class ApnsChannelPool {
 
     private final Set<Future<Channel>> pendingCreateChannelFutures = new HashSet<>();
     private final Queue<Promise<Channel>> pendingAcquisitionPromises = new ArrayDeque<>();
+    private final int maxPendingAcquisition;
 
     private boolean isClosed = false;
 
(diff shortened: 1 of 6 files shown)


jchambers:
Hello, and thank you for the contribution! While I think this is a reasonable proposal for addressing the issue you raised in #1094, I do not think we should proceed with this change for several reasons:

1. As you've noted, this problem occurs when a caller tries to open lots of connections with an expired SSL certificate. Callers can—and should—already avoid this situation by monitoring and managing expiration for their own certificates.
2. Callers can and should also impose some kind of flow control. Please see https://github.com/jchambers/pushy/wiki/Best-practices#flow-control for details, but in general, callers can use something like a `Semaphore` or a [resilience4j `Bulkhead`](https://resilience4j.readme.io/docs/bulkhead) to limit the number of outstanding push notifications.
3. At startup, it's very reasonable for callers to enqueue a large number of push notifications, each of which would wait for a channel to become available. I think this change could introduce tension where a "max outstanding" value small enough to meaningfully limit memory consumption could also interfere with that initial startup burst, leading to confusion.

While I appreciate the contribution, I don't think this is the right solution. I'll leave this open for discussion for a few days just in case I've missed something, though.

I'll add that there's probably still room for improvement in this situation, though; perhaps we could introduce a connection timeout (I've been resistant to timeouts after notifications have been sent, but this is a different case) or add a mechanism for canceling connection futures. To be clear, I think those would be new and separate efforts, and we should not try to add them as part of this pull request.

Sycamore-M:
One more thing to add: the certificate will become expired during use.

Thank your for the reply! I agree that this is not a suitable solution to the issue. I only considered memory overflow and did not think it through completely.

I will read the certificate validity period at startup to ensure that no requests are accepted after the certificate expires.

I expect there will be other solutions in the future.

jchambers:
@Sycamore-M thanks for both your patience and understanding. It sounds like we're on the same page, so let's close this for now.

PR #1101, opened 2025-03-21
Bump Netty to 4.1.119.Final due to CVE-2025-24970

oatz7:
Fixes issue: https://github.com/jchambers/pushy/issues/1100


jchambers:
Thank you for the contribution!

jchambers:
Tests are failing, but that appears to be due to an issue with an older version of the `cache` GitHub Action. I've addressed that in #1102 and run these tests locally, so I'm going to go ahead and merge this as-is.

PR #1104, opened 2025-03-23
Add support for broadcast push notifications

### LINKED ISSUE #1081: Add support for broadcast push notifications
Sounds like there's some new stuff in iOS 18: https://developer.apple.com/videos/play/wwdc2024/10069/

I haven't yet watched the video, but will update this issue when I better understand the scope of the feature and the work required to support it.

jchambers:
This is (for now!) a work-in-progress draft of support for broadcast push notifications. It turns out it's a pretty big project that involves building a client for an entirely new part of APNs: the channel management service.

When it's done, this will close #1081.


jchambers:
For folks following along: it turns out this is an _enormous_ project. It's certainly not a matter of just adding a couple new fields to payloads. There's an entire new API for managing broadcast channels, and there's an entirely new endpoint for sending broadcast notifications—which are distinct from normal push notifications.

I'll almost certainly subdivide this into multiple pull requests, but we're looking at several thousand lines of changes before this is done.

somuMuthaiah:
Hi Jon, when do we think the support for broadcasting will be out?  We rely on Pushy today to send all our push notifications to iOS, and now we are blocked since pushy does not have support yet. 

wondering if we should fork pushy and add these changes or build a custom lib to interface with APNS from scratch which i dont like since Pushy has been wonderful so far or wait for the updates to roll in, wdyt ?

PR #1122, opened 2025-12-02
maintenance updates - using latest netty, slf4j etc, updates to maven plugins…

Lonzak:
Updated the dependencies to the latest (maintained) versions. Also updated the maven-plugins.

Junittest and maven build passed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,7 +105,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-tcnative-boringssl-static</artifactId>
-                <version>2.0.74.Final</version>
+                <version>2.0.75.Final</version>
             </dependency>
             <dependency>
                 <groupId>com.eatthepath</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -105,7 +105,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-tcnative-boringssl-static</artifactId>
-                <version>2.0.62.Final</version>
+                <version>2.0.74.Final</version>
```
m1ngyuan:
```suggestion
                <version>2.0.75.Final</version>
```

Lonzak:
@jchambers I am manually updating the project for some time now. Would it be possible to do a maintenance release? (cp. https://github.com/jchambers/pushy/discussions/1099)

jchambers:
Thank you for the contribution, but I'll update the dependencies by other means. There are some things that come with some of these version bumps that need a little more attention, and I'll set up dependabot to automate dependency updates after this manual run.

PR #1140, opened 2026-05-27
Remove fast-uuid as a dependency

jchambers:
I think it's time to drop `fast-uuid` as a dependency. It's been integrated into the JDK (hooray!) for a while, and now the stock `UUID` serializers/parsers are _at least_ as fast as what `fast-uuid` could do. There's still some benefit to using `fast-uuid` for folks stuck on older JDK versions, but I think almost everybody has moved on at this point, and more people will benefit from removing `fast-uuid` than from keeping it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,11 +90,6 @@
                 <version>2.0.77.Final</version>
             </dependency>
             <dependency>
-                <groupId>com.eatthepath</groupId>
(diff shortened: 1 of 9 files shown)


petarov:
https://bugs.openjdk.org/browse/JDK-8196334

TIL. Very impressive. 👏

